PR #6, opened 2019-08-14
New queries with JSON syntax

amzoss:


nassibnassar:
Could we use json_extract_path_text() for JSON queries as described in https://github.com/folio-org/ldp/blob/master/README.md?

amzoss:
Oh, sure. Sorry, I modeled off the example in Slack, but I'll change it.

nassibnassar:
I think at this point we can use simple names like circ-detail.sql and delete all of the other versions.  They will still exist in the repository history.  I believe we have decided not to use stored procedures/functions but to create a model query exactly as you have done, with placeholders in the query that need to be changed (as long as that is clearly documented).

nassibnassar:
Something to think about might be 4-space indentation.  The closer we can get to fitting within 80 characters, the more we can potentially do things like viewing two queries side-by-side on a medium sized monitor, which can be useful for development.

Also, would it be easier to figure out a testing approach for queries now, or come back to it later?

amzoss:
Have updated code with json_extract_path_text and more consistent style! Happy to talk testing anytime.

PR #7, opened 2019-08-28
new versions of circ-item-detail, services-usage

amzoss:


nassibnassar:
Requesting:
- Output of copy_numbers as a JSON object.
- Why TEXT('X') rather than simply 'X'?

amzoss:
For Redshift, I had to use "json_extract_path_text" because the version that returns the JSON object (json_extract_path) isn't supported. (https://docs.aws.amazon.com/redshift/latest/dg/json-functions.html)

An explicit TEXT cast was also required for Redshift. If I just used 'X', I got an error later trying to cast to date in lines like "SELECT DATE(start_date) FROM parameters" - the error was something like "cannot cast unknown to date". I could have used 'X'::TEXT instead of TEXT('X'), but I thought the TEXT() syntax might be less confusing for novice SQL developers to edit than the :: syntax. TEXT() wasn't required for item_status_filter, but since I had to do it for the dates I thought it would be good to be consistent.

nassibnassar:
Could you use DATE('X') instead of TEXT('X')?

amzoss:
Sure, DATE() works. I've updated the queries and tested on both pg and Redshift. Both queries should be complete in terms of the fields included.

nassibnassar:
Do we still want to output copy_numbers as a JSON object?

amzoss:
The only useful function I could find in Redshift is the json_extract_path_text() one I'm using now. Is there a different way you'd like me to try?

nassibnassar:
I would suggest just wrapping the JSON fragment string "manually" to make it a JSON object.

amzoss:
Do you mean with something like a "to_json" function, like you have in pg? I'm not seeing anything in Redshift that outputs JSON, just tests it or reads from it. https://docs.aws.amazon.com/redshift/latest/dg/json-functions.html

But if you mean using some regex to process the extracted text further, I was having trouble with that, too. Redshift doesn't allow the 'global' flag for regex, and pg requires it, so I could only make compatible code that would do a single replacement.

nassibnassar:
I was thinking something more like '{ . . . ' || json_extract_path_text(i.data, 'copyNumbers') || '}'

nassibnassar:
Maybe: '{ "copyNumbers": ' || json_extract_path_text(i.data, 'copyNumbers') || '}'

amzoss:
Ooooh, sorry, yes. That makes sense. I was just thinking data type, not the object notation.

nassibnassar:
One thing though is that we need to account for the possibility of json_extract_path_text(i.data, 'copyNumbers') being '' or NULL.  Take a look at NULLIF or CASE expressions to help with this.

amzoss:
Sure, I'll look for nulls/empty strings in copy_numbers. I hope the loans apps won't allow for null dates, but just in case I'll also protect the date transformation functions. I'll make a note for other developers to watch out for that, too.

nassibnassar:
It's best to assume that any attribute could be NULL, other than id and data.  Any attribute that is missing in the folio data will show up as NULL in these tables.

nassibnassar:
COALESCE is the function I meant to give that is most useful for dealing with NULL values.  It lets you specify a default value for an argument that might be NULL.

nassibnassar:
I think we should use the double-colon (::) casting syntax (with all-caps data types) throughout for consistency, since that appears to be the only one that works for both DATE and VARCHAR.  If it's too confusing in the WITH parameter section at the top, you could add spaces around the `::`, e.g. `'2019-01-01' :: DATE`.

Could we also do without the blank lines within a query?  Maybe a comment line could be there instead of the blank line.

amzoss:
New versions have ' :: ' syntax throughout, no blank lines, COALESCE to catch null values. Have a first attempt at a better README.

PR #12, opened 2019-09-17
Basic query testing framework

nassibnassar:
Renamed `circ-detail/` to `circ_detail/`, which is necessary for the test code's package management.


amzoss:
From what I can see in the latest commit, it looks like there's still a typo in TESTING.md. It still says "circ-detail.sql" toward the end of the file.

nassibnassar:
You are right.  Fixed.

amzoss:
Looks great! Thanks!

PR #72, opened 2020-09-17
Add invoice lines adjustments derived table

adoerrer:


sql/derived_tables/invoice_lines_adjustments.sql
```diff
@@ -0,0 +1,54 @@
+DROP TABLE IF EXISTS local.rm_invoice_lines_adjustments;
```
nassibnassar, line 1: The `rm_` prefix should not be included in the committed query.

nassibnassar:
What does this comment mean?  "These fields in adjustments can be locally defined"

nassibnassar:
This query does not appear to work in folio_snapshot.  Maybe something has changed in the schema since we loaded most of the data in folio_test_data.

adoerrer:
> 
> 
> This query does not appear to work in folio_snapshot. Maybe something has changed in the schema since we loaded most of the data in folio_test_data.

At this time there are no rows in invoice_lines. I'll ask some aquistions pros to add some invoices to then retest the query.

adoerrer:
> 
> 
> What does this comment mean? "These fields in adjustments can be locally defined"

As I understand institutions can add their own fields in here but I'll try to get more precise information on that and will then change the comment.

PR #73, opened 2020-09-17
adding early versions of new user expansion tables

amzoss:
incomplete (waiting on additional tables) but should be working as is


sql/derived_tables/users_addresses.sql
```diff
@@ -0,0 +1,67 @@
+DROP TABLE IF EXISTS local.angelazoss_users_addresses;
```
nassibnassar, line 1: The `angelazoss_` prefix should not be used in the PR.

nassibnassar:
It might be good to add comments about why certain sections are commented out.

Issue #100, opened 2020-09-29
Develop derived table contributors


jenmawe:
Metadata Management (MM)

Information:
inventory_instances

    to get id, hrid, and json data for contributors

inventory_contributor_name_types
-join to get the name of the contributor from Id in types table

inventory_contributor_types 
-join to get the type from the contributor Id in json data in inventory_instances.data

Issue #99, opened 2020-09-29
Develop derived table alternative_titles


jenmawe:
Metadata Management (MM)

Information: 
inventory_instances 
- to get id, hrid, and json data for alternativeTitles

inventory_alternative_title_types
-join to get the name of the alternative title from Id in json data in inventory_instances.data

PR #121, opened 2020-10-06
2 new derived tables

jenmawe:
Here are the 2 new derived tables that I did.

1. https://github.com/folio-org/folio-analytics/issues/100. This is the instance_contributors.sql
2. https://github.com/folio-org/folio-analytics/issues/99. This is the instance_alternative_titles.sql

I removed the vacuum and analyze from both of these. I also tried to respect the formatting.

Best,
Jennifer

### DIFF
--- a/sql/derived_tables/instance_alternative_titles.sql
+++ b/sql/derived_tables/instance_alternative_titles.sql
@@ -1,0 +1,25 @@
+DROP TABLE if exists local.alternative_titles
+--Create a derived table for alternative titles from instance records
+
+CREATE TABLE local.alternative_titles AS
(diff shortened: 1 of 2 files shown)


nassibnassar:
Thank you.  I think we need to be more careful with consistency:

Issue #99: alternative_titles
File name: instance_alternative_titles
Table name: alternative_titles

Issue #100: contributors
File name: instance_contributors
Table name: instances_contributors (note the "s")

I will fix these.

nassibnassar:
There are also missing semicolons.  These can be found by testing the script before submitting.

Issue #124, opened 2020-10-07
Develop derived table effective_callno_components


nassibnassar:
Extract nested JSON fields from `$.effectiveCallNumberComponents` in table `inventory_items`, and join tables (?).

PR #130, opened 2020-10-13
Submitting Items Effective Call No. Derived Table

jenmawe:
Hi Nassib,

I'm submitting this on behalf of Laura Wright. This is the items effective call number components table, Issue 124. [](https://github.com/folio-org/folio-analytics/issues/124)

I did clean up some of the spacing and punctuation based on what I've done in previous queries. Let me know if there are other changes.

Best,
Jennifer


nassibnassar:
Fixed misspelling of JSON field `typeId` which resulted in no data for column itm_effective_call_number_name.

PR #180, opened 2020-11-16
Create po_lines derived table

NancyBolduc:
Create derived table for po_lines_locations. Main table is po_lines and joining inventory_locations table.


sql/derived_tables/po_lines_locations.sql
```diff
@@ -0,0 +1,36 @@
+DROP TABLE IF EXISTS local.nancybolduc_po_lines_locations;
+--Create a local table for Purchase Order Line Locations including location quantity and type.
+CREATE TABLE local.nancybolduc_po_lines_locations AS
```
nassibnassar, line 3: Note the table name `local.nancybolduc_po_lines_locations` which probably should be `local.po_lines_locations`.  I will fix this after merging.

PR #213, opened 2020-12-02
removing top-level reports folder, updating README files

amzoss:


nassibnassar:
It looks like the fix to loans_items was made in the same branch; so it got included in the PR.  Do you want to include that commit in this merge?

amzoss:
Yes, that would be great!

amzoss:
(If it's okay. Otherwise I can revert.)

PR #243, opened 2020-12-23
Updated holdings and item extended derived tables

### LINKED ISSUE #241: Update Item_ext derived table
While working on the shelf list report for MM, it was noticed that there was a property that was left out, discovery_suppress. This is needed for the shelf list report.

@nassibnassar Could you assign this to me?

### LINKED ISSUE #242: Update holdings_ext derived table
While working on the shelf list report for MM, it was noticed that there was a property that was left out, discovery_suppress. This is needed for the shelf list report.

@nassibnassar Could you assign this to me?

jenmawe:
closes #241 
closes #242


sql/derived_tables/holdings_ext.sql
```diff
@@ -21,6 +21,7 @@ WITH holdings AS (
         json_extract_path_text(data, 'receiptStatus') AS receipt_status,
         json_extract_path_text(data, 'retentionPolicy') AS retention_policy,
         json_extract_path_text(data, 'shelvingTitle') AS shelving_title,
+        json_extract_path_text(data, 'discovery_suppress') AS discovery_suppress,
```
nassibnassar, line 24: Should the JSON field here be referenced as `discoverySuppress` rather than `discovery_suppress`?

sql/derived_tables/item_ext.sql
```diff
@@ -38,7 +38,8 @@ WITH items AS (
         json_extract_path_text(data, 'circulationNotes', 'descriptionOfPieces') AS description_of_pieces,
         json_extract_path_text(data, 'status', 'date') AS status_date,
         json_extract_path_text(data, 'status', 'name') AS status_name,
-        json_extract_path_text(data, 'holdingsRecordId') AS holdings_record_id
+        json_extract_path_text(data, 'holdingsRecordId') AS holdings_record_id,
+        json_extract_path_text(DATA, 'discovery_suppress') AS discovery_suppress
```
nassibnassar, line 42: It is probably a good idea to use lowercase for identifiers, i.e. `data` rather than `DATA`.  DBeaver can be configured not to convert them to uppercase, though it is not entirely obvious at first glance how to do it.

nassibnassar (changes requested):

sql/derived_tables/holdings_ext.sql
jenmawe: I ran a test just with discoverySuppress rather than discovery_suppress for items using inventory_items and both worked. I'm not sure why both discoverySuppress and discovery_suppress would work. I just checked the API and it is discoverySuppress. I just made those changes locally. That is so strange that both worked. Sorry about that.

sql/derived_tables/item_ext.sql
jenmawe: I fixed that also in my local. I caught the uppercase one in the holdings but not the item. For some reason I can only figure out all caps or no all caps. :(

sql/derived_tables/holdings_ext.sql
nassibnassar: Maybe `discoverySuppress` is not present in your test data.  In that case, `json_extract_path_text()` will return `NULL` just as if you used `discovery_suppress`.

nassibnassar (approved):

PR #254, opened 2021-01-06
Add runlist.txt to order derived table queries

nassibnassar:
The new `runlist.txt` file lists all of the derived query files in the order they should be run.  The order is significant because some of the queries depend on the result of other queries.


amzoss (approved):
I think this works as a simple solution for ensuring that derived tables with a dependency are updated after the tables they depend on. I think it can move forward as-is.

I definitely see Axel's point about wanting to mark that dependency some way so it doesn't accidentally get reversed, and I think Sharon's suggestions of adding something to the name of the query with the dependency might actually be a nice convention. Queries with, e.g., a "_dep" suffix could all be left at the bottom of the list by convention. (Not a perfect system because there may be dependencies between tables that themselves have dependencies, but that might be uncommon.)
 
In lieu of that, we can rely on the comments header at the top of the derived table SQL files, but that does take some digging and isn't just immediately apparent in this list.

Derived tables shouldn't be updating as rapidly now that we have so many written, but I we may also want to make sure we document for derived table query writers that they have to add new derived tables to this list for the query to be executed automatically, as well as include a note in the comments header for queries with dependencies.

nassibnassar:
Added `deps.txt` as a suggested way of expressing dependencies explicitly in a human-readable and machine-readable form.

PR #256, opened 2021-01-07
Various adjustments to po_instance derived table

nassibnassar:


sql/derived_tables/po_instance.sql
```diff
@@ -8,18 +8,18 @@ SELECT
...
     json_extract_path_text(po_purchase_orders.data, 'metadata', 'createdByUserId') AS created_by,
     json_extract_path_text(configuration_entries.value::json, 'name') AS created_location,
-    po_lines.instance_id AS pol_instance_id,
+    json_extract_path_text(po_lines.data, 'instanceId') AS pol_instance_id,
     inventory_instances.title AS title,
-    po_lines.publication_date AS date_of_publication,
-    po_lines.publisher AS publisher,
-    po_lines.requester AS requester,
-    po_lines.rush AS rush
+    json_extract_path_text(po_lines.data, 'publicationDate') AS date_of_publication,
```
nassibnassar, line 13: I wonder if this column should be named simply `publication_date`?

jenmawe: I agree Nassib. This would follow the guidelines we talked about in terms of using the property name and if that was camel cased then to separate the words with an underscore. In this case, publication_date.

jenmawe (changes requested):
the property publicationDate should be publication_date in the json extract.

jenmawe (approved):

PR #264, opened 2021-01-08
Update RM Item and title readme documentation

LM-15:


LM-15:
update RM item and title readme documentation

jenmawe (approved):

nassibnassar:
Testing these queries on a large data set, `item_count.sql` completed in 21 minutes and `title_count.sql` ran for more than 8 hours.  I will investigate further.  @adoerrer

adoerrer:
> 
> 
> Testing these queries on a large data set, `item_count.sql` completed in 21 minutes and `title_count.sql` ran for more than 8 hours. I will investigate further. @adoerrer

That's very suprising due that item_count mostly relies on the fields used in title_count and have even more tables and fields joined.

nassibnassar:
The problem is that this subquery in the WHERE clause is throwing off the query optimizer:
```
(SELECT "language"
    FROM folio_reporting.instance_languages AS lng
    WHERE lng.instance_id = inst.instance_id
    LIMIT 1)
```
I'm working on fixing it, but I am confused about what this subquery was intended to do.  If there is more than one language associated with an instance_id, it randomly selects one of them.  Is that really what you want to do?  Could someone clarify what was intended by the subquery above?

LM-15:
I think we are trying to get the first language listed.  Below is what I understood from Laura.  I assume the first language listed in the JSON array is supposed to be from the MARC 008.  If it is not possible to take the first, this is an example of where not having access to the MARC data is a problem?


08/20/2020:

  *   Languages:
It is repeatable in the JSON array.  There may not be a primary language for some titles.  But if there is a primary language, it would go first.  If you want all in a particular language, using all of the repeatable information is fine. If you want no overlap in the count, you would use the first language. MARC:  The primary language, if there is one, is recorded in the 008, positions 35-37 (a 3 letter code).    There has to be something in this spot.  If there are more languages, all are entered in a 041.  The 008 language is in first 041 subfield a. (https://www.loc.gov/marc/bibliographic/bd041.html_)

nassibnassar:
Thank you, that clarifies it.  The problem then is in the `instance_languages` derived table, specifically our temporary workaround for extracting arrays which does not currently preserve ordering.  Let me see if I can correct that.

nassibnassar:
@LM-15 @adoerrer @jenmawe I have submitted a fix for this as a PR to LM-15's main branch: https://github.com/LM-15/folio-analytics/pull/8

adoerrer:
> @LM-15 @adoerrer @jenmawe I have submitted a fix for this as a PR to LM-15's main branch: [LM-15#8](https://github.com/LM-15/folio-analytics/pull/8)

That looks good and the 'LATERAL WITH ORDINALITY'-functionality I haven't know before is a very elegant way to have a sorting within languages. Thank you for that insights! 

Another question: Should we change the item query as well? It is using instance languages in the same way.

nassibnassar:
@adoerrer Yes, please, on the item count query.

jenmawe (approved):

nassibnassar (approved):

PR #278, opened 2021-01-19
refactored loans_and_renewals_counts

### LINKED ISSUE #189: Refactor RA/UM loans and renewals counts report query to use derived tables
[Loans and Renewals Counts query](https://github.com/folio-org/folio-analytics/tree/master/sql/report_queries/loans_and_renewals_counts)
[Loans and Renewals Counts JIRA issue in UXPROD](https://issues.folio.org/browse/UXPROD-1915)

amzoss:
will close #189


NancyBolduc (approved):
I am not very familiar with loans and renewals but it looks good to me!

sql/report_queries/loans_and_renewals_counts/loans_and_renewals_counts.sql
```diff
@@ -1,175 +1,168 @@
...
+                parameters) || ' to '::varchar || (
+        SELECT
+            end_date::varchar
+        FROM
+            parameters) AS date_range,
+    li.loan_date::date,
+    li.loan_due_date AS loan_due_date,
+    li.loan_return_date AS loan_return_date,
+    li.loan_status AS loan_status,
+    '1'::int AS num_loans, -- each row is a single loan
```
nassibnassar, line 82: I wondered if this column is useful.  I would think that `sum(num_loans)` is the same as `count(*)`.  I can see it being more user friendly since `sum(num_loans)` would not require creating a new column; or confusing because it does not convey information in the context of this query.

amzoss: I'm imagining Excel users who will not be using database tools to compute this sum. If they know how to use pivot tables, they won't need the count. Otherwise, they may prefer a number they can sum rather than using Excel's COUNT function, which I think may be less intuitive and which can produce problematic results if you use it on a column with blanks.

amzoss:
Thanks, @nassibnassar! Changes merged.

nassibnassar (approved):

PR #288, opened 2021-01-24
Use uppercase for all SQL reserved keywords

nassibnassar:
Previously we have used Postgres-style SQL, which prefers lowercase for data type and function names.

This is a hypothetical PR to discuss whether to use uppercase for all reserved keywords.  AZ raised the question, as recently we have been using lowercase.

### DIFF
--- a/sql/derived_tables/invoice_lines_fund_distributions.sql
+++ b/sql/derived_tables/invoice_lines_fund_distributions.sql
@@ -22,7 +22,7 @@ SELECT
     json_extract_path_text(ff.data, 'code') AS finance_fund_code,
     json_extract_path_text(ff.data, 'name') AS fund_name,
     json_extract_path_text(ff.data, 'id') AS fund_type_id,
-    ft.name AS fund_type_name,
+    json_extract_path_text(ft.data, 'name') AS fund_type_name,
     fund_distribution_value AS fund_distribution_value,
     fund_distribution_type AS fund_distribution_type,
     invoice_line_sub_total AS invoice_line_sub_total,
--- a/sql/derived_tables/invoice_lines_fund_distributions.sql
+++ b/sql/derived_tables/invoice_lines_fund_distributions.sql
@@ -22,7 +22,7 @@ SELECT
     json_extract_path_text(ff.data, 'code') AS finance_fund_code,
     json_extract_path_text(ff.data, 'name') AS fund_name,
     json_extract_path_text(ff.data, 'id') AS fund_type_id,
-    json_extract_path_text(ft.data, 'name') AS fund_type_name,
+    ft.name AS fund_type_name,
     fund_distribution_value AS fund_distribution_value,
     fund_distribution_type AS fund_distribution_type,
     invoice_line_sub_total AS invoice_line_sub_total,
--- a/sql/pg_format.conf
+++ b/sql/pg_format.conf
@@ -30,11 +30,11 @@ keyword-case=2
 
 # Change the case of the data type name. Default is lowercase: 1.
 # Values: 0=>unchanged, 1=>lowercase, 2=>uppercase, 3=>capitalize.
-type-case=1
+type-case=2
 
 # Change the case of the reserved keyword. Default is unchanged: 0.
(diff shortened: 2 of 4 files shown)


nassibnassar:
There appear to be two common approaches to SQL casing:  (1) Postgres-style uses uppercase for reserved keywords except data type and function names; this is common in Postgres (the most popular open source database system) and its derivatives.  (2) Most other SQL dialects use uppercase for all reserved keywords including data type and function names; this approach is seen more often in industry I believe.

I think we should adopt one of these two styles.  I have a slight preference for 1 but would be fine with 2.

This PR would modify the pg_format configuration from 1 (its present state) to 2, in case there is a desire to make the change.  Note that this results in, e.g., `JSON_EXTRACT_PATH_TEXT()`.

jenmawe:
In terms of selecting either 1 or 2, if there are no implications to staying with 1, what we are doing now, then I would prefer to do that. The only reason is that we have so many other changes right now, this seems a little cosmetic right now.

jenmawe (commented):
I left a comment in the main comments section. Is this only cosmetic? If so, I'd prefer to stay with what we do.

nassibnassar:
@jenmawe It is cosmetic, but this is just a pull request to change the pg_format configuration for future work.  There is no need to reformat all of the existing queries, although this can be done with a simple script if we want to do that.

jenmawe (approved):

adoerrer (approved):
> (2) Most other SQL dialects use uppercase for all reserved keywords including data type and function names; this approach is seen more often in industry I believe.

This explains why this looked so familiar to me and I was tending to 2, thank you. But honestly I'm totally fine with each case, if there are other thoughts on that.

nassibnassar:
I should have made this an issue (for discussion) rather than PR, but I happened to have already located the code to make the changes.  However, I am mainly looking for input on which you would prefer.

amzoss (approved):
Yes, I also slightly prefer uppercase, so I am in favor of this change.

Thanks!

lib-eric:
As I am learning or looking things up, I like to try and follow the documentation so I would opt to follow Case 1.

"...write key words in upper case and names in lower case"
https://www.postgresql.org/docs/13/sql-syntax-lexical.html

Datatype lookup table lists as lowercase
https://www.postgresql.org/docs/13/datatype.html

DBeaver, that many people seem to be using, I believe defaults to uppercase keywords.

I cannot speak to if there is a variation on Redshift and the default interface to connect to it (if different).

todolson:
I'm fine with either Postgresql style or the more industry-standard style, maybe a slight preference for the Postgres style. For me, the significant thing is the keywords pertaining to the structure of the statement (i.e. SELECT, FROM, WHERE, LIMIT, etc.) are upper case, which is common to the two styles. Data types and functions in upper case are maybe slightly distracting, but really either convention are fine so long as we are consistent.

nassibnassar:
We appear to be split down the middle.  This is unfortunate, because I myself also am split down the middle: I like both.

I think the strongest argument for 2 is that it helps table and column names stand out clearly as lowercase.  There is serious merit in that.

Some arguments for 1:

- Postgres-based systems support user-defined functions and/or stored procedures, which are not considered reserved keywords.  That means some function/procedure names could be uppercase and others lowercase, which is a bit too subtle.
- For less technical contributors that may not have access to automatic formatting, typing more uppercase manually seems more tedious.
- We are working in the open source and/or Postgres world, where 1 is clearly the convention.
- With both color syntax highlighting usually available and indentation being used, that seems like enough and I even find myself questioning whether we need uppercase for anything.  (Of course I am not going to propose lowercase for everything.)

Without a clear majority for the uppercasers, let us stay with 1 for now.  If there is no objection, I will close this PR and move the commit to a branch in case we need to revisit the decision in the coming weeks.

amzoss:
Closing the PR sounds fine to me.

nassibnassar:
Synchronizing main branches with reverted commit

nassibnassar:
Moved to commit 3d3960e89891132f196d375824160763e6447d03 in branch sql-reserved-keywords-uppercase, for possible future consideration

PR #301, opened 2021-01-26
Update array extraction in po_lines_physical

nassibnassar:


NancyBolduc (approved):
I see that you added ordinality for pol_volumes. Thanks!

PR #303, opened 2021-01-27
Update array extraction in claimed_returned

nassibnassar:


nassibnassar:
This query may need significant refactoring, but I had already started on the changes before I realized that.

amzoss (approved):
Fine as a pre-refactor solution. Refactored version uses derived tables that need updating.

PR #306, opened 2021-01-27
Clarify name of ordinality column

nassibnassar:


adoerrer (approved):
The added field prefixes should serve a better distinguishment between different ordinalities. Thank you!

PR #321, opened 2021-01-28
Shelf list and refactored pick list

### LINKED ISSUE #222: Refactor MM-Pick List Report
review [Pick List report](https://github.com/folio-org/folio-analytics/tree/master/sql/report_queries/pick_list)

### LINKED ISSUE #224: Create MM-Shelf List Report
There is the shelf list cluster. See [MM working spreadsheet](https://docs.google.com/spreadsheets/d/1QjYa7zy77Ctiys-5MPj70EWqMluSAOnE3b66UH5pnFY/edit#gid=889298050). [UXPROD-2051](https://issues.folio.org/browse/UXPROD-2051?jql=text%20~%20%22Cluster%22%20AND%20%22Report%20Functional%20Area(s)%22%20%3D%20%22Metadata%20Management%22)

jenmawe:
Closes #224 
Closes #222 

I tested both of these on FOLIO sandbox. For the shelf list report, because data changed, I tested both with the filter set to OCLC and then ISBN to ensure that it worked in both cases. However, I didn't test with a large LDP only sandbox. For the pick list, there were some concepts that I'm still wrangling with. I added in the derived tables when I saw it made sense. I might have to take another look at this once I have a better understanding of it.


ameliasutton (commented):
For the Shelf List Changes:

Comments list call no prefix, call no, and call no suffix seperately, table has them combined as eff_call_no

Item Id, Holding Id, and Instance Id seem to be missing from the output

Totally subjective comment: Maybe it would be a good idea to list the actual names of the output fields in the output table in the comments so that there is no ambiguity about which field is in which column. Alternatively just listing the fields included in the order that the output returns them would effectively do the same thing.

jenmawe:
@ameliasutton That's strange. The comments list item effective call number and not the components. The results have the item uuid and hrid and instance id as well.

ameliasutton (commented):
Please ignore my previous comments, I mixed up the two queries when testing them. 

**For the Pick List:**

I do not get results that include the Item Id, Instance Id, or Holdings Id.

The Call No fields are listed separately and not together

**For the Shelf List:**

Missing fields from resulting table (based on the comments):

- Instance id
- Value of Identifier
- Holdings Id
- Holdings hrid


There is only one listed 'discovery_suppression' field and I am not sure if it is for the holdings, instance, or item

Loan Item Status is not listed in the comments but is returned in the report

'location_name' is returned in the table but it is either not listed in the comments or it is unclear which location name data point it is referring to. The 'permenent_location_name' and 'temporary_location_name' columns exist separately

sql/report_queries/pick_list/pick_list.sql
```diff
@@ -97,11 +90,11 @@ instance_format_names AS (
 ),
 identifiers_extract AS (
     SELECT
-        inst.id AS instance_id,
-        JSON_EXTRACT_PATH_TEXT(JSON_ARRAY_ELEMENTS(JSON_EXTRACT_PATH(inst.data, 'identifiers')),'identifierTypeId') :: VARCHAR AS type_id,
-        JSON_EXTRACT_PATH_TEXT(JSON_ARRAY_ELEMENTS(JSON_EXTRACT_PATH(inst.data, 'identifiers')),'value') :: VARCHAR AS value
+        inst.instance_id AS instance_id,
+        inst.identifier_type_id AS type_id,
+        inst.identifier_type_name AS value
```
adoerrer, line 95: I assume `inst.identifier_type_name` is supposed to be `inst.identifier`

sql/report_queries/shelf_list/shelf_list.sql
```diff
@@ -0,0 +1,143 @@
...
+	ii2.identifier_type_name IN('OCLC')
+AND 
+	loc_fil.institution_name = (SELECT institution_filter FROM parameters)
+OR 
+	loc_fil.campus_name = (SELECT campus_filter FROM parameters)
+OR 
+	loc_fil.library_name = (SELECT library_filter FROM parameters)
+OR
+	loc_fil.location_name = (SELECT location_filter FROM parameters)
+;
```
adoerrer, line 143: To get this work I'll suggest: 

```
	(loc_fil.institution_name = (SELECT institution_filter FROM parameters) OR 
		((SELECT institution_filter FROM parameters) = ''))
AND 
	(loc_fil.campus_name = (SELECT campus_filter FROM parameters) OR 
		((SELECT campus_filter FROM parameters) = ''))
AND 
	(loc_fil.library_name = (SELECT library_filter FROM parameters) OR 
		((SELECT library_filter FROM parameters) = ''))
AND
	(loc_fil.location_name = (SELECT location_filter FROM parameters) OR 
		((SELECT location_filter FROM parameters) = ''))
```

Or you might also bring it back to the subquery `location_filtering` where it originally was placed but then you have to change` LEFT JOIN` into `RIGHT JOIN`  in line 121

adoerrer (changes requested):

sql/report_queries/pick_list/pick_list.sql
jenmawe: Thanks for catching that Axel. I just made that change in the file.

sql/report_queries/shelf_list/shelf_list.sql
jenmawe: Thanks again Axel. I updated my report with a commit with this.

adoerrer (approved):

nassibnassar (approved):

Issue #268, opened 2021-01-13
Create template for README.md or new model query


nassibnassar:
We used to look at the circ detail query to see how to structure README.md query documentation.  As an alternative to circ detail which is no longer maintained as a model query, it would be helpful either to select another existing query as a model query or to create a template README.md and/or SQL file.

amzoss:
Does the README format for these two queries work as a recommendation?
- https://github.com/folio-org/folio-analytics/tree/main/sql/report_queries/acrl/circulation
- https://github.com/folio-org/folio-analytics/tree/main/sql/report_queries/services_usage

nassibnassar:
Yes, I think those are good.  Do you propose one of them as a model query or as the basis for a template?

nassibnassar:
Also could we try to have a "maintainer" for each report query, maybe something like:
```
Maintainer: @amzoss
```

nassibnassar:
The "maintainer" idea adds a layer of bookkeeping that is already approximated via Git's history and blame features; and I do not know how practical it is for individual contributors to project their future availability.  I was thinking of it as an optional field but maybe we should leave it out.  I would welcome more thoughts on this.  In general, the less boilerplate the better, given the lack of committed resources.

nassibnassar:
For example, GitHub suggests PR reviewers based on the blame command, which is pretty accurate.  Adding a maintainer field that needs to be updated manually will mean that it is likely to become out of date very easily.

amzoss:
I would propose a simple template:

- Purpose of Report
- Sample of Output
- Query instructions (e.g., how to set parameters, any gotchas)

nassibnassar:
@amzoss I agree with your proposed template.

PR #323, opened 2021-01-28
Create README.md for orders query

sbeltaine:


nassibnassar (commented):
Should we follow the format suggested in https://github.com/folio-org/folio-analytics/issues/268 ?

PR #324, opened 2021-01-28
Suggested changes to invoice adjustment derived tables

nassibnassar:


adoerrer (approved):
The suggested changes are making sense, thank you!

NancyBolduc (approved):
Thanks for making all these changes. Using ratio instead of perc_ makes sense.

PR #348, opened 2021-02-10
Use json_extract_path_text() to get renewal count from circulation_loans

wafschneider:
Fixes an issue in `loans_items` derived table creation if there are no renewals.


wafschneider:
Closing this PR, as it is against the wrong branch 🙄

PR #349, opened 2021-02-10
Use json_extract_path_text() to get renewal count from circulation_loans

wafschneider:
Fixes an issue with derived table `loans_items` when a system has no renewals. Based on `stable` branch (`0.9-release` does not have a `renewal_count` column in the `loans_items` table).

### DIFF
--- a/sql/derived_tables/loans_items.sql
+++ b/sql/derived_tables/loans_items.sql
@@ -74,7 +74,7 @@ SELECT
     iltp.name AS permanent_loan_type_name,
     json_extract_path_text(ii.data, 'temporaryLoanTypeId') AS temporary_loan_type_id,
     iltt.name AS temporary_loan_type_name,
-    cl.renewal_count
+    json_extract_path_text(cl.data, 'renewalCount')::int8 AS renewal_count
 FROM
     circulation_loans AS cl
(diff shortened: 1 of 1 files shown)


nassibnassar:
In the future we should probably make PRs on main or release branches rather than stable.  The documentation about this in CONTRIBUTING.md is wrong; I will fix it.

In this case it does not matter because stable and main are being synchronized during the feature freeze.

nassibnassar:
Also we try not to make PRs from this repository but from forks.

sql/derived_tables/loans_items.sql
```diff
@@ -74,7 +74,7 @@ SELECT
     iltp.name AS permanent_loan_type_name,
     json_extract_path_text(ii.data, 'temporaryLoanTypeId') AS temporary_loan_type_id,
     iltt.name AS temporary_loan_type_name,
-    cl.renewal_count
+    json_extract_path_text(cl.data, 'renewalCount')::int8 AS renewal_count
```
nassibnassar, line 77: We are using `bigint` for `renewalCount`.  I can make this change if needed.

nassibnassar (changes requested):

wafschneider:
Sorry for failing to follow the contributor guidelines! Closing this PR and opening a new one against `main` (with the request datatype change).

PR #351, opened 2021-02-10
updated loc filtering in shelf list

### LINKED ISSUE #350: MM Shelf List Report Location Filtering
It was pointed out that the logic was not right for the location filtering in the shelf list report. This needs to be revised.

jenmawe:
closes #350


amzoss (approved):
Looks good to me! I'm definitely still open to a different way of indicating "anything is fine," but this seems safer for now.

jenmawe:
> 
> 
> Looks good to me! I'm definitely still open to a different way of indicating "anything is fine," but this seems safer for now.

Thanks Angela for looking at this. I tried several variations and with the snapshot data this works well. Another difference from what I did before is that all variables can be null or have no values. Before, I had to have at least one value.

jenmawe (commented):
@nassibnassar I did one last edit on this file. In addition to updating the location filter, I added the identifier as a parameter. This made sense here and I think the report is much better.

nassibnassar (approved):

PR #370, opened 2021-02-26
Fixes #360  added documentation on erm inventory cost query

### LINKED ISSUE #360: Add documention to ERM inventory cost query

adoerrer:


nassibnassar:
From lib-eric:

Can't confirm the output since there is data missing from the "snapshot" and in the "query_dev" there is the data but it does not have the derived tables required for this query.

First parameter "invoice_approval_date" is confusing since the parameters in the SQL is called "start_date" and "end_date". But this might just be because of the workaround they are using right now.

I think the documentation should also mention that the inv.approval_date is being used in place of the "payment_date" for now. If you read the SQL it does mention that but if already mentioning the "paymentDate" in the Future Update, I think it would be nice to mention it there too.

Also I would think "examples" would be better fitting then "options".

Overall I like the idea of the documentation and I think it is useful.

nassibnassar:
I think this query is probably being tested with the database `folio_test_data`.

adoerrer:
Thank you @lib-eric for your review and your thoughts! 
This query can be tested against `folio_snapshot_20210114_invoice` but you have to modify the code slightly to `... local.axeldoerrer_invoice_adjustments_ext AS inv_adj ON invl.invoice_id = inv_adj.invoice_id ...` in line 122 in order to run it. 

The unspecific calls of the dates might make sense as long as there will be only one date that is relevant for a query but I'll take this back to our group to discuss that.

nassibnassar (approved):

Issue #364, opened 2021-02-24
Create invoices_ext derived table


sbeltaine:

adoerrer:
Should be renamed into invoice_transactions derived table

adoerrer:
Has been solved already in the sense of having the needed transaction amounts

PR #377, opened 2021-03-12
Added derived tables fixes #376 and #364

### LINKED ISSUE #376: Create derived table for transactions with order context

adoerrer:


nassibnassar:
Should this PR update runlist.txt?

adoerrer:
> Should this PR update runlist.txt?

Yes, thank you for the reminder. I now added them to runlist.txt and modified invoice_transactions.sql to run even if there are no invoices.

lib-eric:
Testing with our system, running into issues with 'purchase_order_transactions' when trying to run line 31 (pol.description AS pol_description,).

This might just be due to our new refresh. We have had issues where the if a field has not been populated in any record, it does NOT get added to the tables. Or we just do not have complete data yet.

No indexes over MS Access's limit of 32 - so good!

Limited knowledge of finance and funds side but from looking at the query it looks good as far as I can tell.

nassibnassar:
On the issue with `purchase_order_transactions`, it sounds like that might possibly result from the `purchaseOrderTransactions` JSON field not being included in the data extracted from a FOLIO instance.  If so, that is a known issue with LDP that we will try to address at some point in Metadb.  For now we have a standard workaround which is to replace the column references in the query with `json_extract_path_text() AS ...`, which returns `NULL` if the field is missing.

adoerrer:
Thanks, have bow added the workaround for pol_description as well.

PR #379, opened 2021-03-17
Fixes #344

### LINKED ISSUE #344: Add Holdings and Items discovery_suppress to the Title Count and Item Count queries
Add Holdings discovery_suppress to the Title Count query.
Add Holdings discovery_suppress and Items discovery_suppress to the Item Count query.
Should be added as harcoded filters, same as Instance discovery_suppress.

adoerrer:
Added Holdings and Items discovery_suppress to the Title Count and Item Count queries and updated query documentation

### DIFF
--- a/sql/report_queries/erm/count/README.md
+++ b/sql/report_queries/erm/count/README.md
@@ -10,7 +10,7 @@ This query relies on the [title count query](../../title_count) but is only addr
 
 * Each instance has a holdings record.  Each holdings record has a permanent location.
 * Excludes suppressed instance records (instance discovery suppress value is “true”)
-* [When this field becomes available:] Excludes instance records that do not have at least one unsuppressed holdings record (all holdings discovery suppress values are “true”)
+* Excludes instance records that do not have at least one unsuppressed holdings record (all holdings discovery suppress values are “true”)
 * This query is intended to only include e-resources. It includes instance records with instance format names of “computer – online resource” or instance records with holdings library names of “Online”. These values many need to be updated for your local needs.
   </details>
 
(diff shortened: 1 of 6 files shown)


LM-15 (commented):
Hi Axel-
I would remove any text about "at least one unsuppressed holdings record" (because they don't have to worry about this - sorry I added earlier) and "suppress values are true" (because that doesn't seem to be what is used?) .  So each would be:
Excludes suppressed instance and holdings records (for title counts)
OR
Excludes suppressed instance, holdings and item records (for item counts)

as appropriate.

I'm guessing the "inst.discovery_suppress ISNULL" should not be needed in real life, but it doesn't hurt.

LM-15:
Not sure I'm using this correctly Nassib.  Linda

adoerrer:
> Hi Axel-
> I would remove any text about "at least one unsuppressed holdings record" (because they don't have to worry about this - sorry I added earlier) and "suppress values are true" (because that doesn't seem to be what is used?) . So each would be:
> Excludes suppressed instance and holdings records (for title counts)
> OR
> Excludes suppressed instance, holdings and item records (for item counts)
> 
> as appropriate.
> 
> I'm guessing the "inst.discovery_suppress ISNULL" should not be needed in real life, but it doesn't hurt.

Thank you Linda! I've added your suggestions to the PR now. And `inst.discovery_suppress ISNULL` is there in case this field has not been set.

nassibnassar (changes requested):
Could we use the more standard `IS NULL` rather than `ISNULL`?

adoerrer:
> Could we use the more standard `IS NULL` rather than `ISNULL`?
Thank you, have now made that changes.

LM-15 (approved):

nassibnassar (approved):

PR #385, opened 2021-03-19
Changed query/table name

adoerrer:
... for invoices_transactions and purchase_order_transactions** including now module name


NancyBolduc (approved):
Name changed only, so looks good. Great idea to add finance to the name, it will make it easier to find. The run order list seems accurate.

nassibnassar (approved):

Issue #366, opened 2021-02-24
RA/UM create derived table for renewal dates on loans


amzoss:

amzoss:
This is using either LDP history of circulation_loans or the FOLIO loan history endpoint

amzoss:
See notes on https://issues.folio.org/browse/UXPROD-2018

PR #394, opened 2021-03-29
adding derived table to retrieve renewal actions and their dates from loan history

amzoss:
have two versions - one for sql folder and one for sql_2g


amzoss:
will close #366

jenmawe (approved):
I reviewed the loans_renewal_dates.sql

PR #397, opened 2021-03-30
Update po_instance.sql

natalyapik:
added "created_by" from user_users table


jenmawe (approved):

jenmawe:
@nassibnassar This update just added the created_by that was taken out from the users_group table. I ran it through snapshot and it works as expected.

nassibnassar (approved):

PR #401, opened 2021-03-31
updating users_departments_unpacked.sql with now-available department fields 

### LINKED ISSUE #308: RA/UM update users_departments_unpacked
now that the user_departments table appears in the LDP, we can uncomment the lines requiring this table. The table is empty, so everything will have to come from json_extract_path_text, but at least it will be ready whenever that table has data.

amzoss:
can be tested on folio_snapshot_20210329

will close #308


jenmawe (approved):
I ran this first in snapshot and then snapshot_20210329. It ran with no issues but was empty. I think this is just on my end where the data is being filtered out.

nassibnassar (approved):

PR #414, opened 2021-04-15
Creating fund_expenditure_summary.sql , fund directory, and README.md

sbeltaine:


sql/report_queries/fund/fund_expenditure_summary.sql
```diff
@@ -0,0 +1,93 @@
...
+        ''::VARCHAR AS expense_class_filter,
+        /* enter transaction type format as 'payment','pending payment', or 'payment,pending payment' */
+        ''::VARCHAR AS transaction_type_filter
+)
+
+SELECT
+	fitrin.transaction_from_fund_id AS fund_id,
+	fitrin.transaction_from_fund_name AS fund_name,
+	finfuntyp.name AS fund_type_name,
+	fitrin.transaction_from_fund_code AS fund_code,
```
adoerrer, line 45: May I suggest to change these 4 lines into:
```
	finfun.id AS fund_id,
	finfun.name  AS fund_name,
	finfuntyp.name AS fund_type_name,
	finfun.code  AS fund_code,
```
We should take these values from finance_funds because when we have no relating transactions for a fund this values will stay empty.

adoerrer (changes requested):
Looks great but there is one change I might suggest. See the comment for lines 42-45

sbeltaine (commented):
Updated query

adoerrer (approved):
Perfect, thank you!

scotttperry (approved):

PR #417, opened 2021-04-15
New script to add optional columns in public schema

nassibnassar:
This is an idea for a workaround of the missing column problem.


nassibnassar:
Closed this PR to try a different idea.

PR #418, opened 2021-04-15
New file for configuring optional columns in LDP

nassibnassar:


adoerrer (approved):
I appreciate this a lot. This will really help on the missing column issue when no data is present.

amzoss (approved):
Overall I like it! It's lightweight and presumably faster than updating a query when a field is missing.

Two thoughts. One is that I'm not sure about "optional column" for the name. I know naming is always hard and I'm fine with it, but I don't think it's easy to understand right away what this functionality is doing. Something that makes a bit more sense to me is more like "columns_ignore_missing" or something like that, but I'm really fine just going with this.

The other thought I have is that I wonder if there should be additional guidance about selecting the column type. Should we suggest that they check the query that is failing, in case that query expects a certain column type and tries to do some kind of operation on the column? I think this PR works fine as is, but that might be something to consider for the Admin Guide.

Issue #423, opened 2021-04-27
RA/UM - build Exceptions report query


amzoss:
https://issues.folio.org/browse/UXPROD-2401

PR #436, opened 2021-04-27
item-related exceptions report

amzoss:
relevant data seems to come and go in folio_snapshot. We've had luck with the `folio_snapshot_20210414` database, but other dates might also work.

### DIFF
--- a/sql/report_queries/item_related_exceptions/README.md
+++ b/sql/report_queries/item_related_exceptions/README.md
@@ -1,0 +1,28 @@
+# Item-Related Exceptions Report
+
+## Purpose of report
+
+This report shows a list of exceptions, which are actions taken by an operator (staff member) and are recorded in the circulation log. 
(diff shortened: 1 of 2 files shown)


amzoss:
will close #423

lib-eric (approved):
README.md
Looks good. I especially appreciate the link to the documentation in the "Query instructions" section. Still takes a bit of time to read the JSON and figure out the 'actions' but they are there!

Is there a forseeable issue with linking to the documentation, such as the URL for it changing or being out dated?


item_related_exceptions.sql
Was worried for a bit about the repeating lines while using the 'folio_snapshot_20210414_circlog'. Looked at the 'locations_service_points' and saw the ids are the same and so when pulling in location_name you get multiple for that service point and the explains the duplicate lines.

Tested parameter filters - worked as expected.

Well done!!

jenmawe (approved):

PR #442, opened 2021-05-03
Report erm_agreement_package_content_item_list created

stdombek:
The report shows e-resources that are covered by an agreement. All identifiers for a title are aggregated and output in the identificator attribute.

### DIFF
--- a/sql_metadb/report_queries/erm/erm_agreement_package_content_item_list.sql
+++ b/sql_metadb/report_queries/erm/erm_agreement_package_content_item_list.sql
@@ -1,0 +1,51 @@
+/*
+ * description:
+ * -------------
+ * The report shows e-resources that are covered by an agreement.
+ * All identifiers for a title are aggregated and output in the identificator attribute.
+ * 
+ * tables: 
+ * -------------
+ * folio_derived.agreements_package_content_item
+ * folio_derived.agreements_subscription_agreement_entitlement
+ * folio_agreements.work
+ * folio_agreements.erm_resource
+ */
+SELECT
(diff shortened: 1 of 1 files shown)


sql_metadb/report_queries/erm/erm_agreement_package_content_item_list.sql
```diff
@@ -0,0 +1,51 @@
...
+    erm_pci_list.pt_platform_name,
+    erm_pci_list.ti_date_monograph_published,
+    erm_pci_list.ti_first_author,
+    erm_pci_list.ti_monograph_edition,
+    erm_pci_list.ti_monograph_volume,
+    erm_pci_list.ti_first_editor
+ORDER BY
+    "title",
+    "agreement",
+    "res_name"
```
adoerrer, line 51: I  think a semicolon is missing at the end.

adoerrer (changes requested):
Should we add documentation for this report in README.md?

adoerrer (changes requested):
Should we add some documentation?

sql_metadb/report_queries/erm/erm_agreement_package_content_item_list.sql
stdombek: The query can be executed without a semicolon. But to ensure compatibility, I will add the symbol.

stdombek:
If necessary, I create documentation.

adoerrer:
> If necessary, I create documentation.

Thanks! I think only a brief description what it is for and what results are expected should be sufficient.

PR #446, opened 2021-05-06
Additional attributes for the derived table agreements_custom_property

stdombek:
A few additional attributes for the derived table agreements_custom_property. In particular, the ID for linking to the subscription_agreement table.


stdombek:
Apparently there is a problem with an attribute in invoice_voucher_lines_fund_distributions.sql . Please fix it.

PR #447, opened 2021-05-06
Fix to address the duplication of transaction rows

adoerrer:
Fixed issue that caused transactions to be duplicated because of lacking fiscal_year_id as additional key


scotttperry (approved):

NancyBolduc (approved):
Duplication of transaction_id is gone. Thanks for fixing this issue Axel.

nassibnassar:
@adoerrer should this patch also be applied to the 1.1 release?

PR #448, opened 2021-05-07
Additional attributes for the derived table agreements_custom_property

stdombek:
A few additional attributes for the derived table agreements_custom_property. In particular, the ID for linking to the subscription_agreement table.

### DIFF
--- a/sql_metadb/derived_tables/agreements_custom_property.sql
+++ b/sql_metadb/derived_tables/agreements_custom_property.sql
@@ -1,22 +1,52 @@
+/*
+ * Creates a derived table on agreements custom properties
+ */
 DROP TABLE IF EXISTS folio_derived.agreements_custom_property;
 
--- Creates a derived table on agreements_custom_property and
--- resolves the definition name from erm_agreements_custom_property_definition
 CREATE TABLE folio_derived.agreements_custom_property AS
 SELECT
-    id AS custom_property_id,
-    definition_id AS custom_property_definition_id,
-    cpd.pd_name AS custom_property_definition_name,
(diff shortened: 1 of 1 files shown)


stdombek:
Apparently there is a problem with an attribute in derived table invoice_voucher_lines_fund_distributions.sql. This causes the folio-analytics CI / test (pull_request) to fail.

adoerrer (approved):

sql_metadb/derived_tables/agreements_custom_property.sql
```diff
@@ -1,22 +1,52 @@
...
+    agreements_custom_property.public_note AS "custom_property_public_note",
+    agreements_custom_property_definition.pd_id AS "custom_property_definition_uuid",
+    agreements_custom_property_definition.pd_name AS "custom_property_definition_pd_name",
+    agreements_custom_property_definition.pd_type AS "custom_property_definition_pd_type",
+    agreements_custom_property_definition.pd_description AS "custom_property_definition_pd_description",
+    agreements_custom_property_definition.pd_label AS "custom_property_definition_pd_label",
+    agreements_custom_property_integer.id AS "custom_property_integer_id",
+    agreements_custom_property_integer.value AS "custom_property_integer_value",
+    agreements_custom_property_text.id AS "custom_property_text_id",
+    agreements_custom_property_text.value AS "custom_property_text_value"
```
nassibnassar, line 20: If quotation marks ( `"` ) are used around identifiers even (as here) when they are not necessary, be careful that there are no unusual characters accidentally included, because with the quotation marks there Postgres will accept even spaces as part of an identifier name!

PR #469, opened 2021-06-30
ERM-Order Status Report query

### LINKED ISSUE #467: Create ERM-Order Status Report query
[UXPROD-2922](https://issues.folio.org/browse/UXPROD-2922)

adoerrer:
Fixes #467 and can be tested against `folio_marc_test` when commenting out the where conditions


sql/report_queries/erm/erm_order_status.sql
```diff
@@ -0,0 +1,107 @@
...
+    invl.invoice_id,
+    marc.field,
+    string_agg('$'::varchar || marc.sf || marc.content, '') AS series
+FROM
+    po_purchase_orders AS po
+        LEFT JOIN organization_organizations AS vendor ON po.vendor = vendor.id
+        LEFT JOIN po_lines AS pol ON pol.purchase_order_id = po.id
+        LEFT JOIN folio_reporting.instance_ext AS inst ON pol.instance_id = inst.instance_id
+        LEFT JOIN folio_reporting.instance_publication AS pub ON pol.instance_id = pub.instance_id
+        LEFT JOIN instance_authors AS inst_auth ON inst_auth.instance_id = pol.instance_id
```
jenmawe, line 79: I can't see an instance_authors table in FOLIO snapshot public. The only table with authors that I know of is the instance contributors table.

sql/report_queries/erm/erm_order_status.sql
```diff
@@ -0,0 +1,107 @@
...
+        LEFT JOIN instance_ids AS inst_ids ON inst_ids.instance_id = pol.instance_id
```
jenmawe, line 80: In both FOLIO snapshot and marc_test, I just have instance identifiers. There is also the derived table for identifiers in folio reporting. Not sure where instance_ids comes from.

jenmawe (commented):
Once I commented out the where and instance_ids and instance_authors, this ran just fine. To get the authors and instance identifiers, did you want to use the derived tables instance_contributors and instance_identifiers in FOLIO reporting?

sql/report_queries/erm/erm_order_status.sql
adoerrer: I'm sorry, I think my formatting in the WITH clause doesn't make the subqueries not that visible.

I've created a subquery `instance_authors` that put all contributors in `instance_contributors` into one field to prevent duplicating instance rows.

sql/report_queries/erm/erm_order_status.sql
adoerrer: Same here, I've created the subquery `instance_ids` in the WITH-clause and it put all identifiers from `instance_identifiers` in one field to prevent duplicating instance rows.

adoerrer:
> Once I commented out the where and instance_ids and instance_authors, this ran just fine. To get the authors and instance identifiers, did you want to use the derived tables instance_contributors and instance_identifiers in FOLIO reporting?

When running the query with the WITH-clause included it didn't run? At `folio_marc_test` it ran but only one record had `isbns` (identifiers) but nearly every record had `authors` filled.

jenmawe (approved):

amzoss (changes requested):
This all seems to be working just fine! I did some tests of the filters by changing the marc field numbers and the date and things worked as expected.

One thing I would suggest, though, would be to change the name of the parameter field from `days_from_current_date` to `days_prior_to_current_date` or something else that makes it clear that you are subtracting, e.g., 30 days from from the current date. In English, I sometimes read "days from" as "days in the future". 

Just in case you want to make that change quickly, I'll use "request changes." If you prefer to leave it as is, just reply in a comment and I will change my response to approve!

adoerrer:
> This all seems to be working just fine! I did some tests of the filters by changing the marc field numbers and the date and things worked as expected.
> 
> One thing I would suggest, though, would be to change the name of the parameter field from `days_from_current_date` to `days_prior_to_current_date` or something else that makes it clear that you are subtracting, e.g., 30 days from from the current date. In English, I sometimes read "days from" as "days in the future".
> 
> Just in case you want to make that change quickly, I'll use "request changes." If you prefer to leave it as is, just reply in a comment and I will change my response to approve!

Thank you!

That  make sense, I'm happy to change that!

amzoss (approved):
Looks great, thanks!

PR #492, opened 2021-07-27
fixing timestamp conversion in services_usage and loans_items

amzoss:
JSON extract results in a string field in the derived table, so query was failing. fixing both the derived table and the report query, for good measure. Additional slight formatting changes for readability and consistency.

### DIFF
--- a/sql/derived_tables/loans_items.sql
+++ b/sql/derived_tables/loans_items.sql
@@ -29,7 +29,7 @@ SELECT
     cl.loan_date,
     cl.due_date AS loan_due_date,
     cl.return_date AS loan_return_date,
-    json_extract_path_text(cl.data, 'systemReturnDate') AS system_return_date,
+    json_extract_path_text(cl.data, 'systemReturnDate')::timestamptz at time zone 'UTC' AS system_return_date,
     json_extract_path_text(cl.data, 'checkinServicePointId') AS checkin_service_point_id,
     ispi.discovery_display_name AS checkin_service_point_name,
     json_extract_path_text(cl.data, 'checkoutServicePointId') AS checkout_service_point_id,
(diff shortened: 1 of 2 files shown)


sql/derived_tables/loans_items.sql
```diff
@@ -29,7 +29,7 @@ SELECT
     cl.loan_date,
     cl.due_date AS loan_due_date,
     cl.return_date AS loan_return_date,
-    json_extract_path_text(cl.data, 'systemReturnDate') AS system_return_date,
+    json_extract_path_text(cl.data, 'systemReturnDate')::timestamptz at time zone 'UTC' AS system_return_date,
```
nassibnassar, line 32: The addition of `at time zone 'UTC'` converts the `timestamptz` to `timestamp without time zone`.  It would probably be better to leave it as `timestamptz` in a derived table.

nassibnassar, line 32: The addition of `at time zone 'UTC'` converts the `timestamptz` to `timestamp without time zone`.  It would probably be better to leave it as `timestamptz` in a derived table.

nassibnassar (changes requested):

sql/derived_tables/loans_items.sql
lib-eric: Is it worth adding towards the top of the script in the notes something along the lines "By default times are being converted to UTC."?

lib-eric (approved):
Looks great!

Just the thought about adding note about times being in UTC.

sql/derived_tables/loans_items.sql
nassibnassar: @lib-eric  That seems OK to me, but I would add `_utc` to the end of the column name.  On the other hand, if it is kept as `timezonetz`, users can set their local time zone and have it converted automatically.

amzoss: I think `timestamptz` is really the behavior we intended. We intended to match the conversion that LDP does automatically when it detects a column is a timestamp column, and the inclusion of UTC actually seems to give an incorrect result. Thanks for catching that!

nassibnassar (approved):

PR #495, opened 2021-08-22
Port locations_libraries to new schema

nassibnassar:


nassibnassar:
Please disregard the "test commit"

PR #496, opened 2021-09-09
Create README.md

sbeltaine:


adoerrer (approved):
Looks good, thank you!

lib-eric (approved):
Same with this one. When viewing the file it appears you do not need the backslash on the last one `po_lines\` since it appears to show the slash.

Issue #487, opened 2021-07-26
RA/UM - update patron list to add note date


amzoss:
Format of cell should be:
date: note….|date: note…

Issue #488, opened 2021-07-26
RA/UM - review patron list to see why null address type is included


amzoss:
query is bringing in nulls when filtering (ex: use ‘home’ as address type filter and all those with address_type as home, as well as those who have nulls in address type, show up.

amzoss:
this happens when an address type is specified but a set of users have no address with that type. those users are left in the report instead of filtering them out. will add explanatory text in comments to explain this behavior.

PR #498, opened 2021-09-10
updates to patron_list report query

amzoss:
- improved notes list to include date of note
- improved comments to explain address type parameter
- removed direct column references that were preventing the query from running without notes and manual blocks data

best test data is in `folio_snapshot_20210602_users`, but query should also run without errors on snapshot now

### DIFF
--- a/sql/report_queries/patron_list/patron_list.sql
+++ b/sql/report_queries/patron_list/patron_list.sql
@@ -43,16 +43,25 @@ WITH parameters AS (
         ''::varchar AS active_status_filter, -- can be true or false (or '' for either)
         ''::varchar AS is_blocked_filter -- can be true or false (or '' for either)
         ),
-user_notes AS (
+indiv_notes AS (
     SELECT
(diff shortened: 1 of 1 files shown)


amzoss:
will close #487 and #488

lib-eric (approved):
Looks good! Took a bit of tweaking to work for TAMU data but ran eventually.

Good explanation of the address_type_name_filter. Did see duplicates, as noted when not specifying a filter. Other filters worked as expected.

user_notes looks good, works with none, single, and multiple notes.


TAMU doesn't have custom fields so had to remove references to those. Also, I do not see feesfines_manualblocks > data > code but it still ran and results appear to be fine. Probably issue with our older version and/or testing set up.

adoerrer (approved):
Looks good and works well on suggested `folio_snapshot_20210602_users`.

The addressed issues are solved. Results seem to be accurate and filters work as expected. Thanks!

amzoss:
@nassibnassar ready to approve?

PR #500, opened 2021-09-17
remove deprecated fields

### LINKED ISSUE #499: RA/UM - update users_groups to remove deprecated fields
pull created date and updated date from metadata object in users_users since top-level fields are deprecated

amzoss:
will close #499


lib-eric (approved):
Looks good to me!

I see the at the values are slightly different in the `metadata`, is this a more true "source of truth" and that is why it is being changed to this?

amzoss:
We noticed that the non-metadata data elements are marked as deprecated, but we're not sure why they're different!

PR #513, opened 2021-11-04
Add subscription count and subscription cost queries along with readme file

NancyBolduc:


scotttperry (approved):

adoerrer (approved):
Thank you for bringing the subscription count and cost query.

PR #514, opened 2021-11-12
Update po_instance.sql

natalyapik:
added location_id and location name from po_lines table and location name from inventory_locations


lib-eric (approved):

adoerrer (approved):
The query update on adding locations looks good. 
Something to be aware is that this might dupIicate po_lines as there can be multiple locations per  po_line.

PR #515, opened 2021-11-16
Update public.sql

eliana-flo:
Added comments for public.circulation_loans table

### DIFF
--- a/sql/comments/public.sql
+++ b/sql/comments/public.sql
@@ -1,2 +1,22 @@
 -- public.circulation_loans
+COMMENT ON COLUMN public.circulation_loans.id IS 'Unique ID (generated UUID) of the loan';
+COMMENT ON COLUMN public.circulation_loans.user_id IS 'ID of the patron the item was lent to. Required for open loans, not required for closed loans (for anonymization).';
+COMMENT ON COLUMN public.circulation_loans.item_id IS 'ID of the item lent to the patron';
+COMMENT ON COLUMN public.circulation_loans.item_effective_location_id_at_check_out IS 'The effective location, at the time of checkout, of the item loaned to the patron.';
+COMMENT ON COLUMN public.circulation_loans.loan_date IS 'Date time when the loan began (typically represented according to rfc3339 section-5.6. Has not had the date-time format validation applied as was not supported at point of introduction and would now be a breaking change)';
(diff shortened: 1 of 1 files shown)


sql/comments/public.sql
```diff
@@ -1,2 +1,22 @@
...
+COMMENT ON COLUMN public.circulation_loans.renewal_count IS 'Count of how many times a loan has been renewed (incremented by the client)';
+COMMENT ON COLUMN public.circulation_loans.loan_policy_id IS 'ID of last policy used in relation to this loan';
+COMMENT ON COLUMN public.circulation_loans.checkout_service_point_id IS 'ID of the Service Point where the last checkout occured';
+COMMENT ON COLUMN public.circulation_loans.patron_group_id_at_checkout IS 'Patron Group Id at checkout';
+COMMENT ON COLUMN public.circulation_loans.claimed_returned_date IS 'Date and time the item was claimed returned for this loan';
+COMMENT ON COLUMN public.circulation_loans.overdue_fine_policy_id IS 'ID of overdue fines policy at the time the item is check-in or renewed';
+COMMENT ON COLUMN public.circulation_loans.lost_item_policy_id IS 'ID of lost item policy which determines when the item ages to lost and the associated fees or the associated fees if the patron declares the item lost.';
+COMMENT ON COLUMN public.circulation_loans.checkin_service_point_id IS 'Comment to last action performed on a loan';
```
adoerrer, line 21: I assume since this is an ID column it can't be a comment.

adoerrer (changes requested):
Except the line 21 all comments looks good. Thank you for helping on that task!

sql/comments/public.sql
amzoss: agreed, seems like the comment for checkin_service_point_id should be "ID of the Service Point where the last checkin occurred"

amzoss (changes requested):
looks good except for line 21

nassibnassar:
@eliana-flo I think if you make more commits in this branch, for example to address the suggestions above, this pull request should automatically pick up the new commits and add them here, since they will be in the same branch.

eliana-flo:
> @eliana-flo I think if you make more commits in this branch, for example to address the suggestions above, this pull request should automatically pick up the new commits and add them here, since they will be in the same branch.

Thanks Nassib, I made another commit fixing the line! Sorry for not being a better proof reader :)

amzoss (approved):
looks great!

PR #523, opened 2022-01-17
Request documentation of commits and pull requests

nassibnassar:
This clarifies in the contribution guidelines that commits and pull requests should be documented.  Pull request descriptions should be self-contained and descriptive enough to be included in the merge commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,9 @@ cases, manual formatting is preferred.
 -----------------
 
 Names of tables and columns should be in lowercase and use underscores
-( `_` ) to separate words.
+( `_` ) to separate words.  Double underscores ( `__` ) should be
+avoided because they are used in Metadb databases as special
+indicators.
 
 In the rare case where a column name is both the same as a reserved
 word and stands alone without a table or alias prefix, it must be
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,6 +62,11 @@ fixes will be backported to recent release branches.
 A pull request is reviewed by at least two people other than the
 contributor before it can be merged.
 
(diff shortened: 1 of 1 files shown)


amzoss (approved):
These changes are helpful, thanks!

adoerrer (commented):
Does that mean when solving Github issues with a commit or pull request a simple `Fixes #nnn` is not sufficient? 

So in that case you could copy-paste the description from the issues as a description, I assume.

nassibnassar:
@adoerrer Yes, the intent of this proposed change would be to have readable, descriptive commit messages.  Messages like `Fixes #nnn` would be fine to include but preferably not in the short title.  (I will clarify this in the proposed text.)

I think in some cases the issue description could be copied-and-pasted, but sometimes issues in this repository are only three or four words and do not provide any explanation or context.

Now that these queries are being used in production deployments, it will be helpful to document changes in the commits.

nassibnassar:
Here is an example of a merge commit description:

```
    Merge branch 'fund-distribution' into main
    
    Pull report on ERM agreements by cancellation date from Stefan Dombek:
    
    "The report contains an overview of agreements and their cancellation
    deadlines.
    
    Originally, the requirements were to add the data from the order and
    invoice too.  After discussing this, we decided to split the report
    into 3 parts.  The part with the agreements can be found here.  The
    other parts (order and invoice data) are created as separate derived
    tables."
```

adoerrer:
Thank you @nassibnassar for clarifying my question about issue references and make the requested detail level more clear.

adoerrer (approved):
This guidance for commit and pull request documentation gives a clear idea what to include, thank you.

amzoss (approved):

PR #529, opened 2022-01-27
Update public.sql

eliana-flo:
Added comments for the columns of the public.circulation_requests table and added the proxy_user_id column comment for public.circulation_loans.

### DIFF
--- a/sql/comments/public.sql
+++ b/sql/comments/public.sql
@@ -15,8 +15,31 @@ COMMENT ON COLUMN public.circulation_loans.renewal_count IS 'Count of how many t
 COMMENT ON COLUMN public.circulation_loans.loan_policy_id IS 'ID of last policy used in relation to this loan';
 COMMENT ON COLUMN public.circulation_loans.checkout_service_point_id IS 'ID of the Service Point where the last checkout occured';
 COMMENT ON COLUMN public.circulation_loans.patron_group_id_at_checkout IS 'Patron Group Id at checkout';
+COMMENT ON COLUMN public.circulation_loans.proxy_user_id IS 'ID of the user representing a proxy for the patron';
 COMMENT ON COLUMN public.circulation_loans.claimed_returned_date IS 'Date and time the item was claimed returned for this loan';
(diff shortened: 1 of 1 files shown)


sql/comments/public.sql
```diff
@@ -15,8 +15,31 @@ COMMENT ON COLUMN public.circulation_loans.renewal_count IS 'Count of how many t
...
 COMMENT ON COLUMN public.circulation_loans.claimed_returned_date IS 'Date and time the item was claimed returned for this loan';
 COMMENT ON COLUMN public.circulation_loans.overdue_fine_policy_id IS 'ID of overdue fines policy at the time the item is check-in or renewed';
 COMMENT ON COLUMN public.circulation_loans.lost_item_policy_id IS 'ID of lost item policy which determines when the item ages to lost and the associated fees or the associated fees if the patron declares the item lost.';
 COMMENT ON COLUMN public.circulation_loans.checkin_service_point_id IS 'ID of the Service Point where the last checkin occurred';
 
+--public.circulation_requests
+COMMENT ON COLUMN public.circulation_requests.id IS 'UUID of the request';
+COMMENT ON COLUMN public.circulation_requests.cancellation_additional_information IS 'Potential relevant information regarding a cancellation';
+COMMENT ON COLUMN public.circulation_requests.cancellation_reason_id IS 'The id of the request cancellation reason';
+COMMENT ON COLUMN public.circulation_requests.cancelled_by_user_id IS 'The id of the user that cancelled the request';
```
lib-eric, line 28: Capitalize the _ids_  for consistency?

lib-eric:
I like that the `circulation_requests` are ordered as they are in the database, `circulation_loans` is not. That is a little hard to follow but it does not look like it is part of this PR. The consistency would be nice but just having some information is better than none!

Also I question if the `data` column needs to have a comment. I think it is self explanatory "JSON representation of the record".

lib-eric (approved):
Re-ordered and updated capitalization for the IDs. Looks good, thank you!

amzoss (approved):
great new additions, thanks!!

PR #530, opened 2022-02-01
Adding item_ext derived table to Metadb

lib-eric:
-Created item_ext derived table for Metadb.
-Updated runlist to include this table.


nassibnassar:
@lib-eric Apologies; please do remove the `folio_derived` schema prefix from all SQL commands in this PR under `sql_metadb/derived_tables/`, if you would.

nassibnassar:
@lib-eric I think only the `folio_derived` schema prefix should be removed; but not for other schemas.

adoerrer (approved):
Results and code are looking good, thank you!

Issue #516, opened 2021-12-28
RM/ERM - Create derived table: po_line_fund_distribution_transactions.sql


stdombek:
Create a derived table of fund distribution in purchase orders. The derived table contains the information on the fund distribution from the orders app as well as from the transactions system table. The table is required to create reports with a GUI in reporting software solutions.

Database: MetaDB

stdombek:
Derived table is created and merged.

PR #540, opened 2022-02-04
Create derived table po_line_fund_distribution_transactions

stdombek:
A derived table about the fund distribution in the purchase order lines with additional information on financial and transaction data. #516


sql_metadb/derived_tables/po_line_fund_distribution_transactions.sql
```diff
@@ -0,0 +1,166 @@
+DROP TABLE IF EXISTS folio_derived.po_line_fund_distribution_transactions;
```
adoerrer, line 1: The leading schema prefix `folio_derived` for derived tables are not necessary anymore and can be deleted anywhere in the query.

sql_metadb/derived_tables/po_line_fund_distribution_transactions.sql
```diff
@@ -0,0 +1,166 @@
...
+CREATE INDEX ON folio_derived.po_line_fund_distribution_transactions (transaction_amount);
+
+CREATE INDEX ON folio_derived.po_line_fund_distribution_transactions (transaction_encumbrance_initial_amount);
+
+CREATE INDEX ON folio_derived.po_line_fund_distribution_transactions (transaction_encumbrance_amount_expended);
+
+CREATE INDEX ON folio_derived.po_line_fund_distribution_transactions (transaction_encumbrance_order_type);
+
+CREATE INDEX ON folio_derived.po_line_fund_distribution_transactions (transaction_encumbrance_subscription);
+
```
adoerrer, line 166: As we very recently agreed we should add VACUUM ANALYZE to all derived tables., which is missing here.

adoerrer (changes requested):
Thanks, query and results are looking good. Just have a few minor changes requested

adoerrer (approved):
Thanks for the changes.

PR #541, opened 2022-02-04
Metadb derived table for instance stat codes

jenmawe:
Porting over derived table for instance statistical codes from LDP to metadb. There is no issue for this pull.


adoerrer:
This derived table has not been added to the runlist yet.

adoerrer (changes requested):
SQL and results are looking good! It's only missing to the [runlist.txt](https://github.com/folio-org/folio-analytics/blob/main/sql_metadb/derived_tables/runlist.txt)

jenmawe:
I removed the quotes around name and added this to the runlist

sql_metadb/derived_tables/instance_statistical_code.sql
```diff
@@ -0,0 +1,49 @@
...
+SELECT
+	stc.instance_id,
+	stc.instance_hrid,
+	stc.statistical_code_id,
+	stc.stat_code_ordinality,
+	sct.statistical_code_type_id,
+	sctt.name AS statistical_code_type_name,
+	sct.code AS statistical_code,
+	sct.name AS statistical_code_name,
+	stc.stat_code_ordinality
```
amzoss, line 26: This is a duplication of line 21

amzoss (changes requested):
Unfortunately there is a duplication between lines 21 and 26 that prevents this from running for me. Deleting line 26 made it work fine, though!

sql_metadb/derived_tables/instance_statistical_code.sql
jenmawe: This isn't a duplication. There's the name for the code type and then also the code.

jenmawe:
> Unfortunately there is a duplication between lines 21 and 26 that prevents this from running for me. Deleting line 26 made it work fine, though!

Angela, I noticed one other mistake that I'm going to fix. The file should be instance_statistical_codes rather than instance_statistical_code. This is in keeping with our naming convention.

adoerrer (approved):

nassibnassar:
I am going to add a cast for statistical_code_id to uuid.  Also for statistical_code_type_id to uuid, because the JSON transform is not recognizing UUIDs at this point.

jenmawe:
> I am going to add a cast for statistical_code_id to uuid. Also for statistical_code_type_id to uuid, because the JSON transform is not recognizing UUIDs at this point.

Thanks Nassib. I'm trying to be good about the uuids.

amzoss:
Whoops, yes, missed those!

PR #544, opened 2022-02-08
Use numeric(19,4) instead of numeric(12,2) or money for monetary data in Metadb-based queries

nassibnassar:
LDP 1.x tables and queries used the numeric(12,2) data type for monetary data.  For Metadb-based queries we have used both the numeric(12,2) and money types, but going forward we will use numeric(19,4) for monetary data.


stdombek (approved):

stdombek (approved):

nassibnassar:
Apologies, I made a mistake above.  In Postgres, decimal is an alias for numeric, and numeric(19,4) is preferable to decimal(19,4).  This is fixed in: https://github.com/folio-org/folio-analytics/pull/544/commits/378e68840985deb6d45ea8593b112659c4d4bd73

Issue #525, opened 2022-01-24
Port instance_alternative_titles derived table to Metadb


nassibnassar:
The original [instance_alternative_titles](https://github.com/folio-org/folio-analytics/blob/main/sql/derived_tables/instance_alternative_titles.sql) derived table is for LDP 1.x.  Porting it to Metadb might look something like
```sql
SELECT
    id,
    jsonb_extract_path_text(titles.jsonb, 'alternativeTitle') AS alternative_title,
    jsonb_extract_path_text(titles.jsonb, 'alternativeTitleTypeId')::uuid AS alternative_title_type_id,
    titles.ordinality
FROM
    folio_inventory.instance
    CROSS JOIN LATERAL jsonb_array_elements(jsonb_extract_path(jsonb, 'alternativeTitles')) WITH ORDINALITY
        AS titles (jsonb);
```
With the addition of the HRID and (via a join to inventory alternative title types) the type name.

jenmawe:
Thanks for the help on this one. I also solved the issue that I was having with an array with no subproperties. I took inspiration from the LDP instance_editions derived table. Here

editions.data #>> '{}' AS edition,
    editions.ordinality AS edition_ordinality

I incorporated something similar in my test metadb derived instance ext table.

PR #547, opened 2022-02-10
New derived table inst alt titles

jenmawe:
New metadb derived table for instance derived tables. I also added this to the runlist.txt file.


jenmawe:
closes #525

sql_metadb/derived_tables/instance_alternative_titles.sql
```diff
@@ -0,0 +1,31 @@
...
+DROP TABLE IF EXISTS instance_alternative_titles;
+
+CREATE TABLE instance_alternative_titles AS 
+SELECT 
+	i.id AS instance_id,
+	jsonb_extract_path_text(i.jsonb, 'hrid') AS instance_hrid,
+	jsonb_extract_path_text(titles.jsonb, 'alternativeTitle') AS alternative_title,
+    jsonb_extract_path_text(titles.jsonb, 'alternativeTitleTypeId')::uuid AS alternative_title_type_id,
+    titles.ordinality AS alternative_title_ordinality,
+    jsonb_extract_path_text(att.jsonb, 'name') AS alternative_type_name
```
lib-eric, line 13:
```suggestion
    jsonb_extract_path_text(att.jsonb, 'name') AS alternative_title_type_name
```
To match with name in INDEX and to be consistent with naming.

lib-eric (changes requested):
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
- [ ] Column names make sense and are spelled correctly
	See line comment/suggested change (line 13 update select field name)

For derived tables:

sql_metadb/derived_tables/instance_alternative_titles.sql
jenmawe: Thanks Eric. I just made that change.

lib-eric (approved):

PR #561, opened 2022-02-16
Adding port po_lines_details_subscription derived table to Metadb 

### LINKED ISSUE #534: RM - port po_lines_details_subscription to Metadb

adoerrer:
will fix #534

### DIFF
--- a/sql_metadb/derived_tables/po_lines_details_subscription.sql
+++ b/sql_metadb/derived_tables/po_lines_details_subscription.sql
@@ -1,0 +1,22 @@
+DROP TABLE IF EXISTS po_lines_details_subscription;
+
+-- Create a local table for subscription details in po_lines.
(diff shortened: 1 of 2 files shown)


jenmawe (changes requested):
This didn't run. On line 11, should this be folio_orders.purchase_order AS po?

jenmawe (approved):
It runs perfectly. Thanks

lib-eric (approved):
The results of `pol_subscription_interval` looks odd in some places, like where the FROM and TO dates are a year apart but the interval is 1. Or no TO date and interval is 365. Looked at the data from `po_line` table and that is what is in the JSON - so I will chalk it up to limited/bad test data.

- [x] Query runs
- [?] Data output makes sense
- [x] Data types look correct

For derived tables:

PR #566, opened 2022-02-17
Adding port po_lines_physical derived table to Metadb

### LINKED ISSUE #537: RM - port po_lines_physical to Metadb

adoerrer:
will fix #537


jenmawe (approved):
Everything runs well. I checked for double quotes and spelling. Everything looks good.

amzoss (changes requested):
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
- [] Column names make sense and are spelled correctly
For derived tables:
- [x] table appears in runlist.txt at an appropriate point, with correct file name

I believe there is a copy-paste error on line 26 with the name for the material type. This derived table is all about physical inventory, but the field name here is "pol_er_mat_type_name" instead of "pol_phys_mat_type_name". This is also an error in the LDP version of this derived table. 

I know it might break things to change it now, so I understand if we don't want to do that, but I think it would be better to have it correct.

PR #567, opened 2022-02-17
 Adding port po_ongoing derived table to Metadb

### LINKED ISSUE #538: RM - port po_ongoing to Metadb

adoerrer:
will fix #538

### DIFF
--- a/sql_metadb/derived_tables/po_ongoing.sql
+++ b/sql_metadb/derived_tables/po_ongoing.sql
@@ -1,0 +1,28 @@
+DROP TABLE IF EXISTS po_ongoing;
+
+-- Create a derived table for ongoing details in purchase orders.
+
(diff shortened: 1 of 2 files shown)


jenmawe (approved):
This runs and I couldn't find anything of note like use of double quotes. Looks great.

sbeltaine (approved):
Review checklist:
- [yes ] query runs
- [yes] data output makes sense
- [yes] data types look correct
- [yes] query logic seems correct
- [yes] column names make sense and are spelled correctly
- [yes] Query syntax and style are appropriate (e.g., correct use of double quotes, lower case for data types

For derived tables:
- [yes] query has indexes on columns
- [yes]  vacuum analyze has been added to the bottom
- [yes ] table appears in runlist.txt at an appropriate point

PR #568, opened 2022-02-22
new metadb derived table Holdings statements

natalyapik:
add new metadb derived table for holdings also added to runlist

### DIFF
--- a/sql_metadb/derived_tables/holdings_statements.sql
+++ b/sql_metadb/derived_tables/holdings_statements.sql
@@ -9,7 +9,7 @@ SELECT
     h.instanceid AS instance_id,
     h.id AS holdings_id,
     jsonb_extract_path_text(h.jsonb, 'hrid') AS holdings_hrid,
-    jsonb_extract_path_text(hs.jsonb, 'statement') AS "statement",
+    jsonb_extract_path_text(hs.jsonb, 'statement') AS holdings_statement,
     jsonb_extract_path_text(hs.jsonb, 'note') AS public_note,
     jsonb_extract_path_text(hs.jsonb, 'staffNote') AS staff_note,
     hs.ordinality AS statement_ordinality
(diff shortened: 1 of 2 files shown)


jenmawe (changes requested):
The only thing I saw was that in the Create Index at the bottom you have note instead of public_note. Once I changed that, the query ran.
- [] Query runs
- [x] Data output makes sense
- [x] Data types look correct
- [] Column names make sense and are spelled correctly

For derived tables:

adoerrer (approved):
After the update query looks good, thank you! 

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

jenmawe (changes requested):
On line 42, just a change from note to public_note

jenmawe (approved):

sql_metadb/derived_tables/holdings_statements.sql
```diff
@@ -0,0 +1,48 @@
...
+WITH holdings AS (
+SELECT 
+    h.instanceid AS instance_id,
+    h.id AS holdings_id,
+    jsonb_extract_path_text(h.jsonb, 'hrid') AS holdings_hrid,
+    jsonb_extract_path_text(hs.jsonb, 'statement') AS holdings_statement,
+    jsonb_extract_path_text(hs.jsonb, 'note') AS public_note,
+    jsonb_extract_path_text(hs.jsonb, 'staffNote') AS staff_note,
+    hs.ordinality AS statement_ordinality
+    FROM folio_inventory.holdings_record h
```
nassibnassar, line 16: We are tending to use `FROM <table> AS <alias>`.  I will make this change.

sql_metadb/derived_tables/holdings_statements.sql
```diff
@@ -0,0 +1,48 @@
...
+    hs.instance_id,
+    jsonb_extract_path_text(i.jsonb, 'hrid') AS instance_hrid,   
+    hs.holdings_id,
+    hs.holdings_hrid,
+    hs.holdings_statement,
+    hs.public_note,
+    hs.staff_note, 
+    hs.statement_ordinality
+    FROM holdings AS hs
+    LEFT JOIN folio_inventory.instance i ON hs.instance_id = i.id
```
nassibnassar, line 29: And again here.  I will make the change.

PR #569, opened 2022-02-22
Holdings statements index

natalyapik:
added new metadb holdings_statements_indexes table, also added to runlist

### DIFF
--- a/sql/derived_tables/holdings_statements_indexes.sql
+++ b/sql/derived_tables/holdings_statements_indexes.sql
@@ -25,4 +25,4 @@ CREATE INDEX ON holdings_statements_indexes (public_note);
 CREATE INDEX ON holdings_statements_indexes (staff_note);
 
 VACUUM ANALYZE holdings_statements_indexes;
-
+ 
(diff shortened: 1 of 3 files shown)


jenmawe (commented):
There were no results. I think this is because there are no indexes statements in folio juniper. Is that the case?


- [x ] Query runs
- [? ] Data output makes sense
- [x ] Data types look correct
- [ x] Query logic seems correct
- [x ] Column names make sense and are spelled correctly

For derived tables:
- [x ] vacuum analyze has been added to the bottom
- [ x] table appears in runlist.txt at an appropriate point

jenmawe (approved):

adoerrer (approved):
Once I added a test data set the query runs pertfectly, thank you! 

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

nassibnassar (changes requested):
The file name does not match the file name added to runlist or the query name.  Which name would you prefer to use?

nassibnassar:
The file name problem turned into a bit of a mess.  I will fix it in the merge.

PR #570, opened 2022-02-22
new metadb table Holdings statements supplements

natalyapik:
create new metadb table holdings_statement_supplement

### DIFF
--- a/sql_metadb/derived_tables/holdings_statements_supplements.sql
+++ b/sql_metadb/derived_tables/holdings_statements_supplements.sql
@@ -28,20 +28,20 @@ SELECT
     LEFT JOIN folio_inventory.instance AS i ON hs.instance_id = i.id
 ;
 
-CREATE INDEX ON holdings_statements_for_supplements (instance_id);
+CREATE INDEX ON holdings_statements_supplements (instance_id);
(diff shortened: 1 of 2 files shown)


jenmawe:
- [] Query runs
- [x] Data output makes sense
- [x] Data types look correct
- [] Query syntax and style are appropriate (e.g., correct use of double quotes, lower case for data types)

For derived tables:

I just noticed that on line 11, 23, and 39 statement has double quotes. I was able to run the query when I removed the quotes. I realize that sometimes DBeaver will have the warning for some words. If you want, you could have holdings_statement instead of just using the word statement which will then avoid any warnings.

jenmawe (changes requested):
statement is in quotes. See comment in conversation.

jenmawe (approved):
Looks great. All checks passed now

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

adoerrer (approved):
Runs and looks great, thank you! 

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

jenmawe (approved):

Issue #551, opened 2022-02-11
Fix datatypes: derived table licenses_license_ext.sql


stdombek:
Data types must be corrected in the table.

Issue #550, opened 2022-02-11
Fix datatypes: derived table agreements_subscription_agreement_org_ext.sql


stdombek:
Data types must be corrected in the table.

PR #578, opened 2022-03-02
Corrected data type and column names

stdombek:
Column names now without quotes and data type for column "license_org_uuid" changed to UUID. The UUID for "license_id" must still be of the data type varchar (app-internal specification). "license_start" and "license_end" remain integer and must later be set to data type timestamptz. This is currently not possible. 
#551

### DIFF
--- a/sql_metadb/derived_tables/agreements_subscription_agreement_org_ext.sql
+++ b/sql_metadb/derived_tables/agreements_subscription_agreement_org_ext.sql
@@ -4,15 +4,15 @@ DROP TABLE IF EXISTS agreements_subscription_agreement_org_ext;
 -- resolves values and labels from erm_agreements_refdata_value for sao_role
 CREATE TABLE agreements_subscription_agreement_org_ext AS
 SELECT
-    sao_id,
-    sao_owner_fk AS subscription_agreement_id,
-    sao_org_fk AS sao_org_id,
+    sao.sao_id,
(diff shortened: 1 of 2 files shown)


stdombek:
#550

amzoss (approved):
In agreements_subscription_agreement_org_ext.sql, I see several ID columns that haven't been converted to UUID. Is that because they join to tables where the IDs are currently stored as varchar or text? That makes sense to me, but as metadb gets more sophisticated with its type conversion, we might have to revisit this derived table.

Otherwise, these changes seem fine. Thanks for updating the style to match new guidelines! I'll approve on the assumption that there is a good reason for leaving the fields in the first query as varchar.

stdombek:
Hi @amzoss , thank you for the review and feedback. In fact, these are app-specific IDs that are only used internally by App Agreements and are also stored directly in the FOLIO database as a varchar data type. Owen commented that for compatibility purposes this will not be changed either. We can only convert the UUIDs that refer to other apps as a uuid data type.

adoerrer (approved):
Works well, thanks!

**For agreements_subscription_agreement_org_ext.sql**

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
For derived tables:


**For licenses_license_ext.sql**

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
For derived tables:

nassibnassar:
Thanks for the excellent pull request description.

PR #579, opened 2022-03-02
MM - Porting over Derived Table Instance ext

jenmawe:
New metadb derived table for instance.ext

### DIFF
--- a/sql_metadb/derived_tables/instance_ext.sql
+++ b/sql_metadb/derived_tables/instance_ext.sql
@@ -1,0 +1,120 @@
+-- Create an extended version of the instance table where the names for identifiers are included
+-- This derived table doesn't include all the properties extracted from the json objects
+-- For example, electronic access, notes, publication, series, or edition statements have their own 
+-- derived tables.
+DROP TABLE IF EXISTS instance_ext;
+
+CREATE TABLE instance_ext AS 
+WITH expanded_inst AS (
+	SELECT 
+		i.id AS instance_id,
+		jsonb_extract_path_text(i."jsonb", 'hrid') AS instance_hrid,
+		jsonb_extract_path_text(i."jsonb", '_version') AS record_version,
+		jsonb_extract_path_text(i."jsonb", 'matchKey') AS match_key,
+		jsonb_extract_path_text(i."jsonb", 'source') AS record_source,
+		jsonb_extract_path_text(i."jsonb", 'discoverySuppress')::boolean AS discovery_suppress,
+		jsonb_extract_path_text(i."jsonb", 'staffSuppress')::boolean AS staff_suppress,
+		jsonb_extract_path_text(i."jsonb", 'previouslyHeld')::boolean AS previously_held,
+		jsonb_extract_path_text(i."jsonb", 'isBoundWith')::boolean AS is_bound_with,
+		jsonb_extract_path_text(i."jsonb", 'instanceTypeId')::uuid AS instance_type_id,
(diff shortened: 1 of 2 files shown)


sql_metadb/derived_tables/instance_ext.sql
```diff
@@ -0,0 +1,122 @@
...
+-- For example, electronic access, notes, publication, series, or edition statements have their own 
+-- derived tables.
+DROP TABLE IF EXISTS instance_ext;
+
+CREATE TABLE instance_ext AS 
+WITH expanded_inst AS (
+	SELECT 
+		i.id AS instance_id,
+		jsonb_extract_path_text(i.jsonb, 'hrid') AS instance_hrid,
+		jsonb_extract_path_text(i.jsonb, '_version') AS record_version,
```
lib-eric, line 12: Do not think `_version` is required.

My understanding is records in `folio_inventory.instance` are always current; other versions are listed in separate `instance_` table.

sql_metadb/derived_tables/instance_ext.sql
```diff
@@ -0,0 +1,122 @@
...
+    	jsonb_extract_path_text(i.jsonb, 'modeOfIssuanceId')::uuid AS mode_of_issuance_id,
+    	moit.name AS mode_of_issuance_name,
+    	jsonb_extract_path_text(i.jsonb, 'administrativeNotes') AS administrative_notes,
+    	jsonb_extract_path_text(i.jsonb, 'catalogedDate')::date AS cataloged_date,
+    	jsonb_extract_path_text(i.jsonb, 'metadata', 'createdDate')::date AS created_date,
+    	jsonb_extract_path_text(i.jsonb, 'metadata', 'updatedDate')::date AS updated_date
+	FROM 
+		folio_inventory.instance  AS i
+   		CROSS JOIN LATERAL jsonb_array_elements(jsonb_extract_path(i.jsonb, 'classifications')) WITH ORDINALITY AS classes (jsonb)
+		CROSS JOIN LATERAL jsonb_array_elements(jsonb_extract_path(i.jsonb, 'physicalDescriptions')) WITH ORDINALITY AS physdescript (jsonb)
```
lib-eric, line 38: This will result in single instance record having multiple rows. Our previous derived extended tables of Instance, Holdings, and Items have been one-to-one record, then separate derived tables for repeatable fields.

I believe these elements should be broken out into their own derived tables if needed.

sql_metadb/derived_tables/instance_ext.sql
jenmawe: I wasn't sure. There are issues with the property generation. I guess if people really want version then they can query it.

sql_metadb/derived_tables/instance_ext.sql
lib-eric: The `administrative_notes` didn't pop-out at me since it is not in the `CROSS JOIN` - but yes, if they can be repeated, they should be in another table.

I think with the way implemented here you will either get no results (hard to tell with sample data) or it will only be one, first or last.

jenmawe: We discussed the use of pipes in one of our meetings. From that discussion, it seemed people preferred the one to one approach. I just updated the pr and removed classification, phys description and admin notes. I added these our spreadsheet.

lib-eric (approved):

nassibnassar:
It would be helpful if we can figure out how to avoid tab characters in these files.  I will remove them here.

PR #580, opened 2022-03-04
MM - Port Over Derived Table Instance format ids for Metadb

jenmawe:
New ported over instance formats derived table for Metadb.

**NOTE**: There is a problem with some of the FOLIO data for format ids. Instead of a uuid, there is text information. I had to treat the uuids here as varchar because of that.

### DIFF
--- a/sql_metadb/derived_tables/instance_formats.sql
+++ b/sql_metadb/derived_tables/instance_formats.sql
@@ -1,0 +1,42 @@
+-- Create derived table for instance formats bringing together the identifier and name 
+-- Note: Because of inaccurate data in FOLIO, instance_format_id is a varchar and the ift.id has to be cast as a varchar.
+DROP TABLE IF EXISTS instance_formats;
+
+CREATE TABLE instance_formats AS
+WITH instances AS (
+    SELECT
+        i.id AS instance_id,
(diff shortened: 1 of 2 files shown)


amzoss (approved):
Looks good! Thanks!

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
For derived tables:

lib-eric (approved):
Yeah, odd about that varchar | uuid. Good catch on that workaround for it

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

sql_metadb/derived_tables/instance_formats.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE TABLE instance_formats AS
+WITH instances AS (
+    SELECT
+        i.id AS instance_id,
+        jsonb_extract_path_text(i.jsonb, 'hrid') AS instance_hrid,
+        instance_format_ids.jsonb #>> '{}' AS instance_format_id,
+        instance_format_ids.ordinality AS instance_format_ordinality
+    FROM
+        folio_inventory.instance i
```
nassibnassar, line 13: Missing AS.  I will add it.

PR #581, opened 2022-03-04
MM Ported Over Derived Table Instance editions for Metadb

jenmawe:
New derived tabled ported over for instance editions


amzoss (approved):
Looks good, thanks!

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

lib-eric (approved):
Neat and simple :)

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

PR #582, opened 2022-03-04
MM Port Over LDP Instance languages to Metadb

jenmawe:
New derived table for instances languages


amzoss (approved):
Looks good (again)! Thanks!

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

lib-eric (approved):
Only thought is maybe changing `instance_language` to `instance_lang_code`. It is the 3 character for language not spelled out but I think it is still understood either way.

Also looked up in FOLIO instances of the language being "|||" since that looks weird. That is how "Undetermined" are defined -- odd but okay.

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

PR #583, opened 2022-03-04
MM Ported Over Instance Nature of content

jenmawe:
New derived table for metadb


amzoss (approved):
Looks good! 

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

lib-eric (approved):
Good example of the importance of the type casting. Good job!

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

PR #586, opened 2022-03-07
RM - porting over derived table po_lines_tags for Metadb

sbeltaine:


amzoss:
Hi @sbeltaine! It looks like you had trouble getting the changes to runlist.txt (#587) included in this pull request. I think you might be able to fix this by merging your [patch-11](https://github.com/sbeltaine/folio-analytics/tree/patch-11) branch into your [patch-10](https://github.com/sbeltaine/folio-analytics/tree/patch-10) branch, which is the one driving this pull request. You could try the merge using this link:

https://github.com/sbeltaine/folio-analytics/compare/sbeltaine:patch-10...sbeltaine:patch-11?expand=1

If you submit and approve that pull request on your account, I think this open pull request on the folio-org account will update with the runlist.txt changes. When that's done, you could close the other pull request (#587).

If that works for this query, you might try it for your other query as well:

https://github.com/sbeltaine/folio-analytics/compare/sbeltaine:patch-12...sbeltaine:patch-13?expand=1

In the future, I wonder if there is a way to edit multiple files on Github without creating two separate branches. I'd be happy to try it out with you sometime, since I'm also curious how it would work.

lib-eric (approved):
runlist.txt in:
#587 
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

adoerrer (approved):
Looks and works great, thanks!

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
For derived tables:
- [x] table appears in runlist.txt at an appropriate point and with correct file name --> in #587

sbeltaine:
Thanks Axel, Angela, and Eric for reviewing this! I completed the patch merges you recommended, Angela. Yes - I would like to step through this process to see if there is a way to edit multiple files on Github without creating two separate branches.

nassibnassar:
@sbeltaine @amzoss To make multiple commits on the same branch, just select the branch from the dropdown list before you edit.

sql_metadb/derived_tables/po_lines_tags.sql
```diff
@@ -0,0 +1,19 @@
+--po_lines_tags
```
nassibnassar, line 1: Missing description: https://github.com/folio-org/folio-analytics/issues/596

PR #588, opened 2022-03-07
RM - port derived table invoice_lines_fund_distributions for metadb

sbeltaine:


jenmawe (changes requested):
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:
- [] table appears in runlist.txt at an appropriate point

lib-eric (approved):
@sbeltaine No change since I last approved.

Did you mean to re-request review from @jenmawe ?

I'm guessing just need to double check the other PR for the runlist update. Little confusing since two different requests.

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
```diff
@@ -0,0 +1,49 @@
+--invoice_lines_fund_distributions
```
nassibnassar, line 1: Missing description: https://github.com/folio-org/folio-analytics/issues/596

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
```diff
@@ -0,0 +1,49 @@
...
+CREATE TABLE invoice_lines_fund_distributions AS
+WITH funds_distr AS (
+    SELECT
+        id AS invoice_line_id,
+        jsonb_extract_path_text(dist.data, 'code') AS fund_distribution_code,
+        jsonb_extract_path_text(dist.data, 'fundId')::uuid AS fund_distribution_id,
+        jsonb_extract_path_text(dist.data, 'distributionType') AS fund_distribution_type,
+        jsonb_extract_path_text(dist.data, 'value')::numeric AS fund_distribution_value,
+        jsonb_extract_path_text(lines.jsonb, 'subTotal')::numeric(12,2) AS invoice_line_sub_total,
+        jsonb_extract_path_text(lines.jsonb, 'total')::numeric(12,2) AS invoice_line_total
```
nassibnassar, line 14: Please note that monetary values for derived tables in Metadb should be `numeric(19,4)`.  I will make the change here.

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
```diff
@@ -0,0 +1,49 @@
...
+DROP TABLE IF EXISTS invoice_lines_fund_distributions;
+
+CREATE TABLE invoice_lines_fund_distributions AS
+WITH funds_distr AS (
+    SELECT
+        id AS invoice_line_id,
+        jsonb_extract_path_text(dist.data, 'code') AS fund_distribution_code,
+        jsonb_extract_path_text(dist.data, 'fundId')::uuid AS fund_distribution_id,
+        jsonb_extract_path_text(dist.data, 'distributionType') AS fund_distribution_type,
+        jsonb_extract_path_text(dist.data, 'value')::numeric AS fund_distribution_value,
```
nassibnassar, line 12: Please note that monetary values for derived tables in Metadb should be `numeric(19,4)`.  I will make the change here.

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
```diff
@@ -0,0 +1,49 @@
...
+        jsonb_extract_path_text(lines.jsonb, 'subTotal')::numeric(12,2) AS invoice_line_sub_total,
+        jsonb_extract_path_text(lines.jsonb, 'total')::numeric(12,2) AS invoice_line_total
+    FROM
+        folio_invoice.invoice_lines AS lines
+        CROSS JOIN jsonb_array_elements(jsonb_extract_path(lines.jsonb, 'fundDistributions')) AS dist(data)
```
nassibnassar, line 17: I'm not sure this is correct.  I will try to fix it.

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
```diff
@@ -0,0 +1,49 @@
...
+    fund_distribution_id AS fund_distribution_id,
+    ff.fund_status,
+    ff.code AS fund_code,
+    ff.name AS fund_name,
+    ft.id AS fund_type_id,
+    ft.name AS fund_type_name,
+    fund_distribution_value,
+    fund_distribution_type,
+    invoice_line_sub_total,
+    invoice_line_total
```
nassibnassar, line 30: It is probably a good idea to qualify column names with the table name or alias when multiple tables are involved.

Also I wonder about the order of these columns.  For clarity we usually place non-array columns from the base tables before array columns, followed by ordinality, though these are not requirements.

nassibnassar:
I think it might be a good idea to spend some more time with this query before merging.

sql_metadb/derived_tables/invoice_lines_fund_distributions.sql
nassibnassar: Correction, this may be OK.

nassibnassar:
@sbeltaine This PR may need more work.

nassibnassar (changes requested):
At least one change is requested in comments

PR #590, opened 2022-03-08
Simplify instance_ext derived table

nassibnassar:
This change removes a CTE that appears to be unnecessary in the instance_ext derived table.


lib-eric (approved):
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

- [x] table appears in runlist.txt at an appropriate point  (already done)

PR #593, opened 2022-03-08
Performance improvements in holdings_ext

nassibnassar:
This change removes a redundant CTE and prefers columns to JSON fields (using ldp_add_column.conf to avoid missing column errors), which in some cases can improve performance.  In this case the running time was reduced by 65%.


jenmawe (approved):

sql/derived_tables/holdings_ext.sql
```diff
@@ -52,11 +27,11 @@ SELECT
     holdings.retention_policy,
     holdings.shelving_title,
     holdings.discovery_suppress,
-    holdings.created_date,
-    holdings.updated_by_user_id,
-    holdings.updated_date
+    json_extract_path_text(holdings.data, 'metadata', 'createdDate') AS created_date,
```
lib-eric, line 30: Cast to date?
```suggestion
    json_extract_path_text(holdings.data, 'metadata', 'createdDate')::date AS created_date,
```

sql/derived_tables/holdings_ext.sql
```diff
@@ -52,11 +27,11 @@ SELECT
...
+    json_extract_path_text(holdings.data, 'metadata', 'updatedByUserId') AS updated_by_user_id,
+    json_extract_path_text(holdings.data, 'metadata', 'updatedDate') AS updated_date
```
lib-eric, line 32: Cast to date?
```suggestion
    json_extract_path_text(holdings.data, 'metadata', 'updatedDate')::date AS updated_date
```

nassibnassar: Oddly enough it looks like created_date is a timestamp updated_date is a timestamptz.

In any case, this appears to affect lots of derived tables; so I have made an issue for it:  https://github.com/folio-org/folio-analytics/issues/594

Thanks for catching this.

PR #595, opened 2022-03-09
MM, Metadb derived table Instance_ext changed Date to timestamptz

jenmawe:
For system created FOLIO fields of cataloged and updated date change casting to timestamptz.

### DIFF
--- a/sql_metadb/derived_tables/instance_ext.sql
+++ b/sql_metadb/derived_tables/instance_ext.sql
@@ -22,8 +22,8 @@ SELECT
     jsonb_extract_path_text(i.jsonb, 'modeOfIssuanceId')::uuid AS mode_of_issuance_id,
     moit.name AS mode_of_issuance_name,
     jsonb_extract_path_text(i.jsonb, 'catalogedDate')::date AS cataloged_date,
-    jsonb_extract_path_text(i.jsonb, 'metadata', 'createdDate')::date AS created_date,
-    jsonb_extract_path_text(i.jsonb, 'metadata', 'updatedDate')::date AS updated_date
+    jsonb_extract_path_text(i.jsonb, 'metadata', 'createdDate')::timestamptz AS created_date,
+    jsonb_extract_path_text(i.jsonb, 'metadata', 'updatedDate')::timestamptz AS updated_date
 FROM
     folio_inventory.instance AS i
     LEFT JOIN folio_inventory.mode_of_issuance__t AS moit ON jsonb_extract_path_text(i.jsonb, 'modeOfIssuanceId')::uuid = moit.id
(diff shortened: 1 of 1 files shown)


lib-eric (approved):
Small change -- thanks!

adoerrer (changes requested):
Thanks you for changing the date type to the more accurate timestampz. The query runs and looks goods, but I'd suggest to create indexes for `title` and `index_title` as well.

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:
- [] query has indexes on columns

nassibnassar:
@adoerrer @jenmawe I think `title` and `index_title` may tend to be large varchars and unsuitable for the default index type.  They should probably be left without an index for now.

jenmawe:
> @adoerrer @jenmawe I think `title` and `index_title` may tend to be large varchars and unsuitable for the default index type. They should probably be left without an index for now.

Both can get really long. Let me remove these from the index.

adoerrer (approved):
Oh, I didn't know  `title` and `index_title` can get that big. Since they are not created atm @jenmawe you don't have to do anything. 
Thanks, for this work!

sql_metadb/derived_tables/instance_ext.sql
```diff
@@ -22,8 +22,8 @@ SELECT
     jsonb_extract_path_text(i.jsonb, 'modeOfIssuanceId')::uuid AS mode_of_issuance_id,
     moit.name AS mode_of_issuance_name,
     jsonb_extract_path_text(i.jsonb, 'catalogedDate')::date AS cataloged_date,
-    jsonb_extract_path_text(i.jsonb, 'metadata', 'createdDate')::date AS created_date,
-    jsonb_extract_path_text(i.jsonb, 'metadata', 'updatedDate')::date AS updated_date
+    jsonb_extract_path_text(i.jsonb, 'metadata', 'createdDate')::timestamptz AS created_date,
```
nassibnassar, line 25: The `createdDate` data do not appear to have a timezone.  It should probably have the `timestamp` data type.

nassibnassar (changes requested):

nassibnassar:
It looks like you changed `updatedDate` instead of `createdDate`.

jenmawe:
> It looks like you changed `updatedDate` instead of `createdDate`.

Sorry about that. I just updated the createdDate. This should be good to go.

nassibnassar:
Almost!  The `updatedDate` field should be changed back to `timestamptz`.

jenmawe:
> Almost! The `updatedDate` field should be changed back to `timestamptz`.

Changed it back.

PR #598, opened 2022-03-18
Update public.sql

eliana-flo:
Added column comments for the following tables:
-public.circulation_loan_history
-public.circulation_request_preference
-public.circulation_request_policies
-public.circulation_patron_notice_policies
-public.circulation_check_ins
-public.circulation_scheduled_notices


adoerrer (commented):
Thank you @eliana-flo , for all the this great work on the column comments! 
I've a question, some of the IDs were mentioned as UUIDs some not. Would it make sense to mention that all of them are UUIDs?

eliana-flo:
> Thank you @eliana-flo , for all the this great work on the column comments! I've a question, some of the IDs were mentioned as UUIDs some not. Would it make sense to mention that all of them are UUIDs?

That makes sense. Are all FOLIO IDs UUIDs except for HRIDs or is it more complicated than that? I just want to make sure I'm not identifying something as a UUID that I shouldn't be.

stdombek (approved):

adoerrer (approved):
Thank you, great work!

PR #601, opened 2022-03-25
Create MetaDB derived table item_electronic_access

lib-eric:
Ran against folio_juniper - got 2 items with electronic access and in this data the URI and details are the same.


jenmawe (approved):
Works great.

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

adoerrer (approved):
looks and works great, thank you!

- [x] Data output makes sense
- [x] Data types look correct
For derived tables:

PR #610, opened 2022-04-20
Create metadb derived table items_holdings_instances

lib-eric:
Create a metadb derived table for linking items to holdings to instance. Table is primarily item level with piece information such as volume and material type, but also has holding level details like call number fields, and at the instance level with title and cataloged date.

### DIFF
--- a/sql_metadb/derived_tables/items_holdings_instances.sql
+++ b/sql_metadb/derived_tables/items_holdings_instances.sql
@@ -11,7 +11,6 @@ SELECT
     ii.chronology,
     ii.enumeration,
     ii.volume,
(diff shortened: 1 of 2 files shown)


jenmawe (approved):
This looks good. The only thing I saw was that the PR didn't have a description. I vaguely remember that Nassib wanted PRs to have descriptions.

- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

adoerrer (approved):
Thanks, looks and work great!
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct
For derived tables:

PR #611, opened 2022-04-20
Create metadb derived table item_tags

lib-eric:
Create a metadb derived table for items with tags. Table contains a new row for every tag associated with an item record and contains the name of the tag and ordinality the tag appears in.

### DIFF
--- a/sql_metadb/derived_tables/item_tags.sql
+++ b/sql_metadb/derived_tables/item_tags.sql
@@ -1,0 +1,21 @@
+-- metadb -- item_tags
+-- This table creates a listing of items and their tag(s).
+
+DROP TABLE IF EXISTS item_tags;
+
+CREATE TABLE item_tags AS
+SELECT
+    i.id AS item_id,
+    tags.data #>> '{}' AS item_tag,
+    tags.ordinality AS item_tag_ordinality
+FROM
+    folio_inventory.item AS i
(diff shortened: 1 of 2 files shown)


stdombek (approved):
- [X] Query runs
- [X] Data output makes sense
- [X] Data types look correct
- [X] Query logic seems correct
- [X] Column names make sense and are spelled correctly
- [X] Query syntax and style are appropriate (e.g., correct use of double quotes, lower case for data types)

For derived tables:
- [X] query has indexes on columns
- [X] vacuum analyze has been added to the bottom
- [X] table appears in runlist.txt at an appropriate point

jenmawe (approved):
- [x] Query runs
- [x] Data output makes sense
- [x] Data types look correct

For derived tables:

nassibnassar (changes requested):
If possible could we have a PR description: https://github.com/folio-org/folio-analytics/blob/main/CONTRIBUTING.md#pull-request-description

jenmawe:
@lib-eric I noticed that you didn't include the item hrid. Did you want to do that? Our other derived tables include both id and hrid.

lib-eric:
> @lib-eric I noticed that you didn't include the item hrid. Did you want to do that? Our other derived tables include both id and hrid.

Good point @jenmawe , added!

stdombek (approved):

adoerrer:
> If possible could we have a PR description: https://github.com/folio-org/folio-analytics/blob/main/CONTRIBUTING.md#pull-request-description

I think this is solved now and the PR is ready to be merged, thanks.

PR #617, opened 2022-05-10
Updated column comments for LDP 1.x 

eliana-flo:
Added column comments for user_users table and added missing column comments for other tables.


stdombek (approved):

adoerrer (approved):
Thank you Eliana for your work on these, looks great!

sql/comments/public.sql
```diff
@@ -12,40 +12,76 @@ COMMENT ON COLUMN public.circulation_loans.item_status IS 'Last item status used
...
 COMMENT ON COLUMN public.circulation_requests.hold_shelf_expiration_date IS 'Date when an item on the hold shelf expires';
 COMMENT ON COLUMN public.circulation_requests.holdings_record_id IS 'ID (generated UUID) of the holdings record being requested';
+COMMENT ON COLUMN public.circulation_requests.instance_title IS 'Title of the item being requested';
 COMMENT ON COLUMN public.circulation_requests.instance_id IS 'ID (generated UUID) of the instance being requested';
+COMMENT ON COLUMN public.circulation_requests.item_barcode IS 'Barcode of the item being requested';
 COMMENT ON COLUMN public.circulation_requests.item_id IS 'ID (generated UUID) of the item being requested';
+COMMENT ON COLUMN public.circulation_metadata_created_by_user_id IS 'ID (generated UUID) of the user who created the record (when available)';
+COMMENT ON COLUMN public.circulation_metadata_created_date IS 'Date and time when the record was created';
+COMMENT ON COLUMN public.circulation_metadata_updated_by_user_id IS 'ID (generated UUID) of the user who last updated the record (when available)';
+COMMENT ON COLUMN public.circulation_metadata_updated_date IS 'Date and time when the record was last updated';
```
nassibnassar, line 47: These appear to be table names rather than column names.

nassibnassar (changes requested):

sql/comments/public.sql
eliana-flo: Whoops! You're right. I goofed up. Looks like I just missed a word/decimal point. I'll take a look and make adjustments.

stdombek:
I'm not entirely sure, but I see 2 underscores in ldp between metadata and the following words in the attribute names.

eliana-flo:
> I'm not entirely sure, but I see 2 underscores in ldp between metadata and the following words in the attribute names.

D'oh! I didn't notice that. Thank you for catching that.

stdombek (approved):
Looks very good.

sql/comments/public.sql
```diff
@@ -79,13 +123,48 @@ COMMENT ON COLUMN public.circulation_check_ins.occurred_date_time IS 'Date and t
...
+COMMENT ON COLUMN public.user_users.patron_group IS 'ID (generated UUID) corresponding to the group the user belongs to';
+COMMENT ON COLUMN public.user_users.personal__date_of_birth IS 'The users birth date';
+COMMENT ON COLUMN public.user_users.personal__email IS 'The users email address';
+COMMENT ON COLUMN public.user_users.personal__first_name IS 'The users given name';
+COMMENT ON COLUMN public.user_users.personal__last_name IS 'The users surname';
+COMMENT ON COLUMN public.user_users.personal__middle_name IS 'The users middle name (if any)';
+COMMENT ON COLUMN public.user_users.personal__mobile_phone IS 'The users mobile phone number';
+COMMENT ON COLUMN public.user_users.personal__phone IS 'The users primary phone number';
+COMMENT ON COLUMN public.user_users.personal__preferred_contact_type_id IS 'ID of users preferred contact type like Email, Mail or Text Message';
+COMMENT ON COLUMN public.user_users.personal__preferred_first_name IS 'The users preferred name';
```
lib-eric, line 165: Pretty insignificant but maybe these need to be `user's` instead of just `users`?

Other then that - looks great to me!

lib-eric (changes requested):

sql/comments/public.sql
stdombek: @nassibnassar Are there processing problems when single quotes there?

eliana-flo: @stdombek I added some double single quotes in so I think it should be okay now.

nassibnassar: I think the escaping here looked correct to me.

lib-eric (approved):

PR #621, opened 2022-05-17
Added fund_detail sql

NancyBolduc:


NancyBolduc:
This query is for LDP. It rovides the list of all active funds along with their ledger, fund group and fund type.  It also provides allocations, transfers, encumbrances, expenditures, fund balance and percentage spent for each fund.

PR #622, opened 2022-05-17
appr_invoices_bib_data

NancyBolduc:
Query for approved invoices with bib data info


NancyBolduc:
This is for LDP

nassibnassar:
From Sharon: we will remove reference to local tables.

PR #638, opened 2022-05-26
MM New Metadb Derived Table for Instance admin notes

jenmawe:
This table extracts administrative notes from the instance. Please note that there is no data for this in juniper.

PR #639, opened 2022-05-26
MM New MetaDB Derived Table Instance tags

jenmawe:
This derived table extracts tags from the instance record. Please note there are no tags in the instance records when this was tested on folio_juniper


nassibnassar:
- [ ] Pull request has a complete description
- [ ] Query output is correct
- [ ] Query is readable and properly indented
- [ ] Quotation marks are used only where necessary
- [ ] File name is listed in `runlist.txt` after dependencies
- [ ] All columns have indexes
- [ ] Table is vacuumed and analyzed

sql_metadb/derived_tables/runlist.txt
```diff
@@ -26,6 +26,7 @@ instance_formats.sql
 instance_editions.sql
 instance_languages.sql
 instance_nature_content.sql
+instance_tags.sql
```
stdombek, line 29: There are no dependencies on other derived tables. So it should be sorted alphabetically. See also discussion in PR #638

adoerrer:
* [x]  Pull request has a complete description
     * [x]  Query runs without errors
     * [ ]  Query output is correct
     * [x]  Query logic is clear and well documented
     * [x]  Query is readable and properly indented
     * [x]  Table and column names are in all-lowercase
     * [x]  Quotation marks are used only where necessary
     * [x]  Query begins with user documentation in comment lines
     * [x]  File name is listed in runlist.txt after dependencies
     * [x]  All columns have indexes
     * [x]  Table is vacuumed and analyzed
 
 
As @stdombek mentioned there is no test data to verify if the output is correct.

@jenmawe I did not find any option to add tags to an instance via UI. It is only possible for holdings and items. How can tags be added to an instance? If this is not possible this derived table might be obsolete.

adoerrer (commented):
Looks good and without errors, but there's no test data to verify its output.

jenmawe:
@adoerrer I thought that the ability to add tags to instances came with Lotus. But I think this might be Morning Glory for instances. The work is done and that's why we can see this in the API documentation. Tags has been implemented both in the back end and UI for holdings and items.

adoerrer (approved):
Thank you, for clarifying this @jenmawe. 
I would propose to approve this query since it does not run into errors.

stdombek (approved):
I agree with @adoerrer.

PR #651, opened 2022-06-15
MM new metadb derived table item_administrative_notes

lib-eric:
This table extracts administrative notes from item records. Must use Lotus for fields to be present.


stdombek (approved):
- [X] Pull request has a complete description
- [X] Query runs without errors
- [X] Query output is correct
- [X] Query logic is clear and well documented
- [X] Query is readable and properly indented
- [X] Table and column names are in all-lowercase
- [X] Quotation marks are used only where necessary
- [X] Query begins with user documentation in comment lines
- [X] File name is listed in runlist.txt after dependencies
- [X] All columns have indexes
- [X] Table is vacuumed and analyzed

PR #657, opened 2022-07-11
Pol vendor ref number

natalyapik:
Create new LDP derived table po_lines_vendor_reference_numbers.


amzoss (changes requested):
This query looks useful, in that it expands an array inside the po_lines table. One thing to note is that if there is no data in the vendorDetail/referenceNumbers field, the query returns an empty table (no rows). 

I would love for the PR poster to add some test data to FOLIO snapshot so we could test it with real data. I think it's still set up so that new test data added earlier in the day will show up later in the LDP, so I'd be happy to coordinating testing on a day when someone can add test data early on. 

Additionally, I wonder if it's better for the derived table to at least show some records from the PO line table, even if they don't have reference numbers. I'm including some sample code below that would do that, but I don't know if it's the best solution. Maybe you do actually want an empty table if there are no reference numbers.

Just some thoughts! But there are also specific errors mentioned in the line comments that prevent the query from running - namely, in lines 3 and 5 you need to remove the schema name. Also, I would love for the po_instance.sql file to be removed from this PR, since those changes are being reviewed in a separate PR.

Thanks!

All queries:
- [ ] Query output is correct  \(not sure, basically a judgment call\)

Derived tables:

Alternate query idea

```CREATE TABLE po_lines_vendor_reference_numbers AS 
WITH ref_numbers AS (
	SELECT
		po_lines.id AS po_line_id,
		json_extract_path_text(numbers.data, 'refNumber') AS vendor_reference_number,
		json_extract_path_text(numbers.data, 'refNumberType') AS vendor_reference_number_type
	FROM po_lines
		CROSS JOIN json_array_elements(json_extract_path(po_lines.data, 'vendorDetail', 'referenceNumbers'))
		AS numbers (data)
)
SELECT
	po_lines.id AS po_line_id,
	po_lines.po_line_number AS po_line_number,
	ref_numbers.vendor_reference_number,
	ref_numbers.vendor_reference_number_type,
	json_extract_path_text(po_lines.data, 'vendorDetail', 'instructions') AS vendor_instructions
FROM po_lines
	LEFT JOIN ref_numbers ON po_lines.id = ref_numbers.po_line_id;
```

nassibnassar:
I am not sure why this PR deletes po_instance.sql.  I will try to undo that.

nassibnassar:
Merging commits manually and closing this PR to avoid deleting po_instance.sql

PR #654, opened 2022-06-28
Update RM po_lines_locations derived table for LDP

sbeltaine:
This update is to add holdings location from po line.

PR #658, opened 2022-07-13
RM - update derived table po_lines_locations  / LDP

adoerrer:
This will add holding_id and relying locations resolving on newly introduced holding information at po line level for derived table po_lines_locations

This PR credits on the original PR #654 from @sbeltaine

### DIFF
--- a/sql/derived_tables/po_lines_locations.sql
+++ b/sql/derived_tables/po_lines_locations.sql
@@ -1,4 +1,4 @@
-/* This query creates a local table for purchase order line locations including location quantity and type.
+/* This query creates a derived table for purchase order line locations including location quantity and type.
 It uses conditional statements to search for location ID and location name in the 
 purchase order line location and holdings location fields, displays whichever is not null,
 and indicates the source of the resulting location data (purchase order location or purchase
@@ -16,7 +16,8 @@ SELECT
     CASE WHEN json_extract_path_text(locations.data, 'locationId') IS NOT NULL
     	THEN json_extract_path_text(locations.data, 'locationId') 
 		ELSE ih.permanent_location_id END AS pol_location_id,	
-	CASE WHEN (il.name) IS NOT NULL THEN il.name
+	CASE WHEN il.name IS NOT NULL
+	    THEN il.name
 		ELSE il2.name END AS pol_location_name,
 	CASE WHEN il.name IS NOT NULL
          THEN 'pol_location'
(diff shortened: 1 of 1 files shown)


sql/derived_tables/po_lines_locations.sql
```diff
@@ -1,40 +1,40 @@
...
+
+SELECT
+	pol.id AS pol_id,
+	json_extract_path_text(locations.data, 'quantity') AS pol_location_qty,
+	json_extract_path_text(locations.data, 'quantityElectronic') AS pol_loc_qty_elec,
+	json_extract_path_text(locations.data, 'quantityPhysical') AS pol_loc_qty_phys,      
+    CASE WHEN json_extract_path_text(locations.data, 'locationId') IS NOT NULL
+    	THEN json_extract_path_text(locations.data, 'locationId') 
+		ELSE ih.permanent_location_id END AS pol_location_id,	
+	CASE WHEN (il.name) IS NOT NULL THEN il.name
```
amzoss, line 19: Slight style inconsistency here - using parentheses around the field name here, but not in line 16 or 21. Also, in other lines, `THEN` is placed on a new line.

amzoss (approved):
I made a couple of style comments, but otherwise this all looks fine to me! It's nice to see a solution without a CTE, and I think the location logic makes sense, especially since it explains the source of the location data.

All queries:
- [x] Query output is correct

Derived tables:

sql/derived_tables/po_lines_locations.sql
```diff
@@ -1,40 +1,40 @@
-DROP TABLE IF EXISTS po_lines_locations;
+/* This query creates a local table for purchase order line locations including location quantity and type.
```
eliana-flo, line 1: I wonder if the description should say "derived table" instead of "local table." In my mind, those are two different things.

eliana-flo (approved):
Looking good! I think the CASE WHEN logic is very elegant and worked well in Snapshot. Made a nitpicky comment about the description, but all the boxes were checked for me. 

All queries:
- [x] Query output is correct

Derived tables:

sql/derived_tables/po_lines_locations.sql
adoerrer: Thank you for catching these!

sql/derived_tables/po_lines_locations.sql
adoerrer: Good catch! Have changed that.

adoerrer:
Thanks @amzoss and @eliana-flo for reviewing and approving the PR. Since I've commited some minor changes you suggested should there be an re-review or should we just let it go?

eliana-flo:
> Thanks @amzoss and @eliana-flo for reviewing and approving the PR. Since I've committed some minor changes you suggested should there be an re-review or should we just let it go?

I think the changes are minor enough that we can let it go. Thanks for the quick work! :)

amzoss (approved):
I'll approve again, just in case! :)

PR #662, opened 2022-08-17
holdings_administrative_notes

natalyapik:
create new LDP derived table holdings_administrative_notes


sql/derived_tables/holdings_administrative_notes.sql
```diff
@@ -0,0 +1,25 @@
...
+
+CREATE TABLE holdings_administrative_notes AS
+SELECT
+    h.id AS holdings_id,
+    h.hrid AS holdings_hrid,
+    admin_note.data #>> '{}' AS administrative_note,
+    admin_note.ordinality AS administrative_note_ordinality
+FROM
+    inventory_holdings AS h
+    CROSS JOIN LATERAL json_array_elements(json_extract_path(data, 'administrativeNotes'))
```
jenmawe, line 14: We want lowercase instead of the camel case here with administrativeNotes.

sql/derived_tables/po_instance.sql
```diff
@@ -1,63 +1,80 @@
-DROP TABLE IF EXISTS po_instance;
+-- Create a local table for purchase order line instance. Every po line will have location ID or holding ID
```
jenmawe, line 1: This looks like the same derived table that was with your PR for instance_administrative_notes. Is that intentional?

sql/derived_tables/po_instance.sql
```diff
@@ -1,63 +1,80 @@
-DROP TABLE IF EXISTS po_instance;
+-- Create a local table for purchase order line instance. Every po line will have location ID or holding ID 
+--depending on how the po is created.
 
-CREATE TABLE po_instance AS
+DROP TABLE IF EXISTS local_core.po_instance;
+
+CREATE TABLE local_core.po_instance AS
```
jenmawe, line 6: This has the schema "local.core" here. Is this your local schema for Cornell?

jenmawe (changes requested):
The title of the derived table has a spelling error, holdigns instead of holdings.

All queries:
- [x] Query output is correct

Derived tables:

jenmawe (commented):
The only thing I noticed was the name of the derived query filename which is holdigns_administrative_notes rather than holdings_administrative_notes.

sql/derived_tables/holdings_administrative_notes.sql
```diff
@@ -0,0 +1,25 @@
+-- This derived table extracts administrative notes from the instance
```
amzoss, line 1: Small error in the documentation. This should say holdings instead of instance.

amzoss (changes requested):
Again, I don't think the camel case in the name for the CROSS JOIN table is a problem, since we have other derived table queries that use camel case there. But there is an error in the comment in the first line that I would ask to fix before approving.

All queries:
- [x] Query output is correct

Derived tables:

sql/derived_tables/holdings_administrative_notes.sql
nassibnassar: It is true that identifiers, such as names of tables and columns, should be in lowercase and should use underscores to separate words, rather than using camel case; this is documented in the contributing guidelines at https://github.com/folio-org/folio-analytics/blob/main/CONTRIBUTING.md#8-naming-things .  However, `'administrativeNotes'` is not an identifier but a string (note the surrounding `'`), and it refers to a JSON field; JSON fields do use camel case, and the camel case should be preserved when referring to them.

amzoss (approved):
Documentation fix looks great, thanks!

All queries:
- [x] Query output is correct

Derived tables:

amzoss:
@jenmawe I think this is ready for you to review again. If you don't have time, should we find another reviewer?

jenmawe (approved):

PR #663, opened 2022-08-17
instance_administrative_notes

natalyapik:
create new LDP derived table instance_administrative_notes


sql/derived_tables/instance_administrative_notes.sql
```diff
@@ -0,0 +1,25 @@
...
+
+CREATE TABLE instance_administrative_notes AS
+SELECT
+    i.id AS instance_id,
+    i.hrid AS instance_hrid,
+    admin_note.data #>> '{}' AS administrative_note,
+    admin_note.ordinality AS administrative_note_ordinality
+FROM
+    inventory_instances AS i
+    CROSS JOIN LATERAL json_array_elements(json_extract_path(data, 'administrativeNotes'))
```
jenmawe, line 14: It's my understanding that we want to avoid the use of camel case. Here you have administrativeNotes. Could you change it to administrative_notes or administrativenotes?

sql/derived_tables/po_instance.sql
```diff
@@ -1,63 +1,80 @@
-DROP TABLE IF EXISTS po_instance;
+-- Create a local table for purchase order line instance. Every po line will have location ID or holding ID 
+--depending on how the po is created.
 
-CREATE TABLE po_instance AS
+DROP TABLE IF EXISTS local_core.po_instance;
+
+CREATE TABLE local_core.po_instance AS
```
jenmawe, line 6: This has your schema local_core which should be left out..

jenmawe (changes requested):
All queries:
- [x] Query output is correct

Derived tables:

jenmawe (approved):

amzoss:
Hi @natalyapik! I know the po_instance.sql file stuff is confusing, and hopefully we can get that figured out after your other PR for po_instance.sql is fixed and approved. For now, though, when you actually delete that file from your branch, it is a bit problematic, because then it's like a request to delete that file from the main repository, too.

Instead of deleting the file from your branches, could you just make sure the file on the branches matches the file on the main repo? So, you could download the file from https://github.com/folio-org/folio-analytics/blob/main/sql/derived_tables/po_instance.sql and add it to the sql/derived_tables folder for this branch and the branch for #662. I think that would be the correct fix here, instead of just deleting the file.

Would you be willing to try that while we're waiting for the second reviews to come in?

natalyapik:
Hi Angela,
I’m more than willing to do that but I think I need help. Po_instance table got screwed up and also did not make it to 1.4 release. I would like to fix that as well. Can we zoom perhaps and doit together? I am fully open today
Natalya.

amzoss (approved):
I would disagree with the other reviewer about changing the camel case on the CROSS JOIN. We have other derived table queries that use camel case just for that one table name, and I don't know if it's worth extra effort to fix all of them, since that table is just temporary and never shown to the users. I personally think this is fine as-is, since there is precedent.

All queries:
- [x] Query output is correct

Derived tables:

sql/derived_tables/instance_administrative_notes.sql
nassibnassar: It is true that identifiers, such as names of tables and columns, should be in lowercase and should use underscores to separate words, rather than using camel case; this is documented in the contributing guidelines at https://github.com/folio-org/folio-analytics/blob/main/CONTRIBUTING.md#8-naming-things .  However, `'administrativeNotes'` is not an identifier but a string (note the surrounding `'`), and it refers to a JSON field; JSON fields do use camel case, and the camel case should be preserved when referring to them.

amzoss: *slaps forehead* Yes, of course! Sorry for adding confusion.

nassibnassar:
Merged manually as https://github.com/folio-org/folio-analytics/commit/f2bb80d712abef4a3d2fbc75e4fb776dbbc916ff

PR #713, opened 2023-01-13
Port feesfines_accounts_actions to Metadb

amzoss:
```
All queries:
- [ ] Pull request is based on a new branch (not main)
- [ ] Pull request scope is not overly broad
- [ ] Release notes have been added or updated in CHANGES.md
- [ ] Query output is correct
- [ ] Query is readable and properly indented with spaces (not tabs)
- [ ] Quotation marks are used only where necessary

Report queries:
- [ ] Query has complete user documentation
    - [ ] Purpose of report
    - [ ] Sample output
    - [ ] Query instructions

Derived tables:
- [ ] File name is listed in `runlist.txt` after dependencies
- [ ] All columns have indexes
- [ ] Table is vacuumed and analyzed
```


eliana-flo (changes requested):
**Some notes:**

1. Should the "fine_date" and "fine_updated_date" columns be formatted to timestamptz like the "transaction_date" column?
2. What's the point of having both the "user_patron_group_id" and the "patron_group_id?" This seems redundant. 
3. We should rename "operator_id" (like you said). Where does the "source" field come from? Maybe we can just call it "operator?"

All queries:
- [x] Pull request is based on a new branch (not main)
- [x] Pull request scope is not overly broad
- [x] Release notes have been added or updated in CHANGES.md
- [] Query output is correct

Derived tables:

amzoss:
Hi @eliana-flo! Sorry for the huge delay!

I wanted to limit the number of changes I propose in this PR, but I know there's a lot of gray area. For example, the LDP1 table hasn't gotten the operator_id change yet, so I figured this PR should just include a straight port from the LDP query, and then I would propose another PR to make the changes to both versions. I think that's the case for point 2, as well - that error is also in the LDP1 query.

For the first point, I'm actually not sure what the data coming from FOLIO look like, since I haven't been able to access a Metadb or LDP1 instance where those fields are getting populated. But the API docs say it's a date-time field, so I agree that either timestamp or timestamptz would make sense.

What do you think, should I correct points 2 and 3 right now? And should I just add in the LDP1 corrections into this PR?

Thanks!

eliana-flo:
> Hi @eliana-flo! Sorry for the huge delay!
> 
> I wanted to limit the number of changes I propose in this PR, but I know there's a lot of gray area. For example, the LDP1 table hasn't gotten the operator_id change yet, so I figured this PR should just include a straight port from the LDP1 query, and then I would propose another PR to make the changes to both versions. I think that's the case for point 2, as well - that error is also in the LDP1 query.
> 
> For the first point, I'm actually not sure what the data coming from FOLIO look like, since I haven't been able to access a Metadb or LDP1 instance where those fields are getting populated. But the API docs say it's a date-time field, so I agree that either timestamp or timestamptz would make sense. I can try to confirm that and make that change here.
> 
> What do you think, should I also correct points 2 and 3 right now? And should I just add in the LDP1 corrections into this PR?
> 
> Thanks!

Hi Angela, I hadn't considered that you would go back and update the LDP1 derived table as well. If that's the case, then I think it makes sense to do updates 2 &3  (LDP1 + Metadb) in a separate PR. Feel free to request a review from me when you do!

amzoss:
No, I definitely could have been more clear about that!

nassibnassar:
If possible, please add
```
--metadb:table feesfines_accounts_actions
```
as the very first line of the file, followed by a blank line.

This will help Metadb to report on progress of loading the derived tables.  See for example:  https://github.com/folio-org/folio-analytics/blob/main/sql_metadb/derived_tables/holdings_notes.sql

eliana-flo (approved):

amzoss:
@nassibnassar I've added the new Metadb directive. @eliana-flo I know you already approved (thanks!) but I did go ahead and cast to timestamptz. In the next PR I'll update both the LDP1 and Metadb tables to use the `metadata` object for the date values and fix a bunch of other things.

nassibnassar (approved):

PR #725, opened 2023-02-27
add Metadb directive to reviewer checklist

amzoss:
Update both contributing guidelines and review checklist with information about new Metadb directives requirement.


nassibnassar:
Applied in commit 88ed395

PR #745, opened 2023-04-18
ACQ/ERM - LDP1 derived table po_lines_fund_distribution

### LINKED ISSUE #744: ACQ/ERM - LDP1 derived table po_lines_fund_distribution
The working groups ACQ and ERM needs a derived table for LDP1 that extracts the data for the fund distribution from the po lines.

stdombek:
Closes #744 
The working groups ACQ and ERM needs a derived table for LDP1 that extracts the data for the fund distribution from the po lines. This table generates the required data.


sql/derived_tables/po_lines_fund_distribution.sql
```diff
@@ -0,0 +1,45 @@
+-- This derived table shows data of the fund distribution of the purchase order lines
+
+DROP TABLE IF EXISTS po_lines_fund_distribution;
+
+CREATE TABLE po_lines_fund_distribution AS
+SELECT
+    po_lines.id AS pol_id,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'fundId') AS pol_distr_fund_id,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'value') AS pol_distr_value,
```
nassibnassar, line 9: It looks like `pol_distr_value` is a number.  If it is a monetary amount, I suggest casting to `numeric(19,4)`.  Otherwise `numeric` might be the best choice.

nassibnassar (changes requested):

sql/derived_tables/po_lines_fund_distribution.sql
nassibnassar: Sorry, I looked at this some more.   If `pol_distr_value` is a percentage, then it should probably be `numeric(19, 2)`.

If `pol_distr_type` determines the type of value in `pol_distr_value`, then I think the `pol_distr_value` should be `numeric`.

nassibnassar (approved):

sql/derived_tables/po_lines_fund_distribution.sql
stdombek: Thank you @nassibnassar for the feedback. The value can be an amount if you selected in the user interface amount or it can be percentage if you selected percentage. This is very differently selected by the user. 
You could do calculations and unify it. On column with the calculated amount and another one for the calculated percentage. However, this calculation brings with it many problems if you have ongoing orders. In the calculation you need the estimate price (or other totals) for 100%, but that value changes with the fiscal year rollover. The fiscal year rollover process overwritten the po lines and also the estimated price. The calculation would be wrong after the rollover. For this reason, this derived table only outputs the value and no calculations. I see no problem in changing the data type. The only question would be with what precision? numeric(19, 4) or numeric(19, 2)?

nassibnassar: If it will **only** be a monetary amount or a percentage, then `numeric(19, 4)` would work for both.  If it could be other kinds of numbers, then just plain `numeric`.  Thanks!

sql/derived_tables/po_lines_fund_distribution.sql
```diff
@@ -0,0 +1,45 @@
...
+
+CREATE TABLE po_lines_fund_distribution AS
+SELECT
+    po_lines.id AS pol_id,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'fundId') AS pol_distr_fund_id,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'value')::numeric(19, 4) AS pol_distr_value,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'distributionType') AS pol_distr_type,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'code') AS pol_distr_code,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'encumbrance') AS pol_distr_encumbrance_id,
+    json_extract_path_text(json_array_elements(po_lines.data->'fundDistribution'), 'expenseClassId') AS pol_distr_expense_class_id
```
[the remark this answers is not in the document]
nassibnassar: @adoerrer Good catch; but LDP does not generally use the uuid type.  At the time I was looking at compatibility with other database systems.

nassibnassar:
A recent change: `CREATE INDEX` and `VACUUM ANALYZE` statements are no longer needed.  I will remove them in this PR while merging it.

nassibnassar:
Once we have a third review, I will merge this together with the duplicate https://github.com/folio-org/folio-analytics/pull/778 .

stdombek:
Also closes #756

VandanaShah-Cornell (approved):
All queries:
- [x] Query output is correct

Derived tables:
- [Not applicable; this is for LDP ] First line is "--metadb:table" directive, followed by blank line

PR #758, opened 2023-05-17
ACQ/ERM - Rpt transactions

### LINKED ISSUE #753: ACQ/ERM - Create Rpt transactions
The report should show the transaction data with their expense classes and finance data.

stdombek:
Closes #753  
There can be many transactions in the database. The report offers the possibility to filter within the transactions according to certain parameters. Therefore the report contains several attributes from the FOLIO finance app.

The report was created for LDP1 and Metadb.


VandanaShah-Cornell (approved):
- [x] Query output is correct

Report queries:
- [x] Query has complete user documentation
    - [x] Purpose of report
    - [x ] Sample output
    - [x] Query instructions

sql/report_queries/transactions/transaction_report.sql
```diff
@@ -0,0 +1,173 @@
...
+    finance_funds.fund_code,
+    finance_funds.fund_name,
+    finance_funds.fund_status,
+    finance_funds.fund_type_name,
+    finance_funds.ledger_code,
+    finance_funds.ledger_name,
+    finance_funds.ledger_status
+FROM 
+    public.finance_transactions 
+    LEFT JOIN public.finance_expense_classes ON finance_expense_classes.id = json_extract_path_text(finance_transactions.data, 'expenseClassId')
```
LauSchl, line 88: Everything seems to work fine. Just one question:
The Checklists asks for a standardized format for extracting json data.
For LDP 1&2 it suggests: t #>> '{f1,f2,f3}'.
This report however uses jsonb_extract_path_text(t, f1, f2, f3) for both ldp and metadb.
I'm just wondering if it will function like that consistantly as well.


All queries:
- [x] Query output is correct
- [ ] JSON extraction is in standard form, for example:

Report queries:
- [x] Query has complete user documentation
    - [x] Purpose of report
    - [x] Sample output
    - [x] Query instructions

LauSchl (commented):
See comment.
If that checks out it's working great :)

sql/report_queries/transactions/transaction_report.sql
nassibnassar: This is a good point.  I think we can be less strict about queries under report_queries/ since they will often have to be adjusted anyway in a specific library.

stdombek: Unfortunately, the queries were created before we had the new regulations and the new reporting app. If requested, I can update the queries according to the new specifications. At the time of the PR, there was no reporting app and the report queries were stored in subfolders. The question is whether the query for the new app could perhaps be completely recreated. 
This time we are so behind on the reviews that we have new regulations again. :-)

LauSchl: Then I see no reason why I schouldn't approve if the queries are going to be adjusted anyway.
Or do you want to change it @stdombek ?

stdombek: We leave this query in this form. If the new variant for the reporting app is available, a separate query is created.

LauSchl (approved):

PR #771, opened 2023-06-09
Change JSON operations in LDP derived tables to work with both json and jsonb types

nassibnassar:
LDP 2.0 is planned to use the `jsonb` data type instead of `json`, which will be a breaking change.  This PR attempts to convert all JSON operations in LDP derived tables to support both data types, in order to minimize the impact of upgrading to LDP 2.0.

Note that these changes use the `#>` and `#>>` JSON operators (and other techniques) to allow us to continue supporting LDP 1.x for some period of time while also supporting LDP 2.0.  However, for Metadb derived tables, I would still recommend standardizing on functions such as `jsonb_extract_path_text()` as they are significantly more readable.

No need to approve this PR, but comments are welcome.


stdombek:
That sounds exciting @nassibnassar . I noticed that when a type conversion took place after the extraction function (json_extract...), the brackets of the function were not deleted. I'm not sure what the implications of this are, or if it's even necessary.

nassibnassar:
> That sounds exciting @nassibnassar . I noticed that when a type conversion took place after the extraction function (json_extract...), the brackets of the function were not deleted. I'm not sure what the implications of this are, or if it's even necessary.

Thanks, yes, the parentheses are required when doing a type conversion, because otherwise the type conversion is evaluated first.

eliana-flo:
Thank you for tagging me in this so I can stay in the loop. I wasn't aware of LDP 2.0. Since Metadb is supposed to take over eventually is 2.0 going to be the last iteration of the OG LDP? Can we keep calling it "LDP1" when trying to distinguish between it and Metadb?

nassibnassar:
> Thank you for tagging me in this so I can stay in the loop. I wasn't aware of LDP 2.0. Since Metadb is supposed to take over eventually is 2.0 going to be the last iteration of the OG LDP? Can we keep calling it "LDP1" when trying to distinguish between it and Metadb?

My guess is that 2.x will be the last versions of LDP, but it is hard to know for sure.

Good question about the naming.  My efforts to fix the naming problems have not succeeded.  But in any case, it might be useful to differentiate between LDP 1 and 2.

PR #778, opened 2023-06-28
Add derived table to extract fund distribution from po_lines

nassibnassar:
Fund distribution is a JSON array in "public.po_lines".  This query extracts the arrays into a table.


stdombek:
What ist the different to PR #745 ?

nassibnassar:
> What is the difference to PR #745 ?

Probably no difference.  Someone asked me today how to extract the array, and I had forgotten about #745 .  I will look at both PRs and merge them together.

stdombek:
That sounds good. Your request is based on the new criteria. So I would suggest taking this one and just adding the comments from the other. :-)

nassibnassar:
Closing, changes moved to PR #788

PR #790, opened 2023-08-09
Backport item_administrative_notes to LDP

nassibnassar:
Backport item_administrative_notes to LDP

### DIFF
--- a/sql/derived_tables/item_administrative_notes.sql
+++ b/sql/derived_tables/item_administrative_notes.sql
@@ -1,0 +1,22 @@
+-- This derived table extracts administrative notes from the item
+-- record.
+
+DROP TABLE IF EXISTS item_administrative_notes;
+
+CREATE TABLE item_administrative_notes AS
+SELECT
+    i.id AS item_id,
+    i.hrid AS item_hrid,
+    i.holdings_record_id AS holdings_record_id,
+    admin_notes.data #>> '{}' AS administrative_note,
+    admin_notes.ordinality AS administrative_note_ordinality
+FROM
+    public.inventory_items AS i
+    CROSS JOIN LATERAL jsonb_array_elements((data #> '{administrativeNotes}')::jsonb)
+        WITH ORDINALITY AS admin_notes (data);
+
+CREATE INDEX ON item_administrative_notes (item_id);
+
+CREATE INDEX ON item_administrative_notes (item_hrid);
+
+CREATE INDEX ON item_administrative_notes (holdings_record_id);
--- a/sql/derived_tables/runlist.txt
+++ b/sql/derived_tables/runlist.txt
@@ -61,3 +61,4 @@ users_groups.sql
 po_prod_ids.sql
 loans_renewal_dates.sql
 locations_service_points.sql
+item_administrative_notes.sql


LauSchl (approved):
works fine :)

All queries:
- [x] Query output is correct

Derived tables:
- [ ] First line is "--metadb:table" directive, followed by blank line

VandanaShah-Cornell (commented):
Question from Joanne Leary at Cornell: Why is this item_administrative_notes query pulling records where there are no administrative notes? Running this on Cornell's ldp, the results have only 24 rows of data, with about half showing [null] in the notes field. If nulls are supposed to show then why aren't the millions of other null records showing up? In the instance_administrative_notes derived table, the same thing happens - in cases where there are multiple notes, one of which is null, the record with the Null administrative note will appear in the results set. However, in cases where there is only a single administrative note, which is Null, and no other notes, the instance will still appear in the results set.

nassibnassar:
@VandanaShah-Cornell I just saw the problem.  Some of the source data have a JSON array containing one element of null:
```
[null]
```
These can be removed with:
```
WHERE administrative_note IS NOT NULL
```

VandanaShah-Cornell:
Hi Nassib,

I wasn’t sure how to reply to a GitHub email, so I’m copying you on your regular email. I’m also copying Joanne because I don’t know enough about the notes field.

Anyway, here is a screenshot of the results of the query (pull request #790). Joanne is puzzled as to why records with null notes are showing up in the results, and I thought that maybe all item records are included irrespective of whether they have notes or not – but in that case, running it on Cornell’s ldp, we would have millions, and not just 24 rows.

***@***.***

Thanks,
Vandana

VandanaShah-Cornell:
I guess we are still wondering why all the other nulls didn’t show up.

nassibnassar:
@VandanaShah-Cornell The null values you are seeing is in the JSON, and I don't know if they might be significant in some way in the FOLIO context, because they are elements in an array.  If you can determine that those nulls have no significance, we can remove them as well in the query.

nassibnassar:
@VandanaShah-Cornell Try this query:
```
SELECT id,
       data #>> '{administrativeNotes}' AS administrative_notes
    FROM public.inventory_items
    ORDER BY length(data #>> '{administrativeNotes}') DESC;
```
It might take several minutes to run.  If you scroll down through the results, you will find administrative notes that look like this:
```
[ null ]
```
And some that look like this:
```
[]
```
The first is an array with one null element.  I'm not sure whether that information might be useful; so the array is preserved.  If the null there has no significance, we can remove the row in the derived table.

The second is an array with no elements.  Since there are no elements, there are no rows for that item in the derived table.

VandanaShah-Cornell:
Thanks, Nassib! That’s what I understood after your last message, so I’m glad I understood it correctly. I will check with Joanne as she is the biggest user of notes.

nassibnassar:
@VandanaShah-Cornell Also, this question can also be asked of all of the derived tables that extract JSON arrays.

VandanaShah-Cornell:
Joanne says the rows where the notes contain just nulls are of no use and should be removed so as to not cause any confusion.

That’s a very good question about all derived tables. I can ask at the next Reporting SIG meeting.

I’m also wondering why we are spending much time on derived tables because once everyone has MetaDB, derived tables won’t be used much, right? Or will lots of institutions stay with the LDP?

nassibnassar:
@VandanaShah-Cornell This probably needs to be confirmed with the FOLIO developers.  For example, do these JSON arrays all mean the same thing?  If we remove those null elements, they will be indistinguishable in LDP/Metadb.
```
[]
```
```
[ null ]
```
```
[ null null ]
```

VandanaShah-Cornell:
That’s a good question for the SIG, guess we should leave it alone for now, and people can filter out the nulls in their query where they include the derived tables.
Thanks so much for clarifying all this. I’ll forward this discussion to the others.

VandanaShah-Cornell (approved):
- [x] Query output is correct

Derived tables:
- [x] First line is "--metadb:table" directive, followed by blank line

PR #849, opened 2024-06-05
fix for Ldp items_holdings_instances.sql derived table

natalyapik:
there is a mistake in pulling instance hrid from instance_holdings table instead it should be pulled from inventory_instances table


natalyapik:
closes issue#850

Issue #856, opened 2024-06-13
Fix Metadb derived table holdings_ext


VandanaShah-Cornell:
1. Fixed to bring in the correct data for the 'updated_by_user_id' field; it was bringing in data from the created by userid rather than the updated_by_user_id. 
2. The holdings_supress field needed to be cast as Boolean. 
3. The json extract for ill_policy_id needed to be cast as UUID.
4. Cast the date fields as "timestamptz" (changed from text)
5. Cast the discovery_suppress field as Boolean (changed from text)
6. Cast the extract of updatedByUserId as UUID (changed from text)

sbeltaine:
Merged

PR #857, opened 2024-06-13
Update holdings_ext.sql

VandanaShah-Cornell:
/*Changes: 
1. Fixed to bring in the correct data for the 'updated_by_user_id' field; it was bringing in data from the created by userid rather than the updated_by_user_id. 
2. The holdings_supress field needed to be cast as Boolean. 
3. The json extract for ill_policy_id needed to be cast as UUID.
4. Cast the date fields as "timestamptz" (changed from text)
5. Cast the discovery_suppress field as Boolean (changed from text)
6. Cast the extract of updatedByUserId as UUID (changed from text) */

### DIFF
--- a/sql_metadb/derived_tables/holdings_ext.sql
+++ b/sql_metadb/derived_tables/holdings_ext.sql
@@ -8,7 +8,6 @@ DROP TABLE IF EXISTS holdings_ext;
 
 CREATE TABLE holdings_ext AS
 
-WITH holdings AS (
     SELECT
         h__t.id,
         h__t.hrid as holdings_hrid,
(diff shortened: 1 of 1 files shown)


VandanaShah-Cornell:
This PR closes issue #856

trapido (changes requested):
There are a couple of problems: 
1.The query won’t run because the CTE is not properly terminated with “)” and the main query is missing a select statement. The easiest way to fix this might be to remove line 11, “WITH holdings AS ( “.
2.I think that joining holdings_record__t and holdings_record tables might be redundant: it should be possible to extract all of the data out of holdings_record table.

All queries:
- [ ] Query output is correct

Derived tables:
- [x] First line is "--metadb:table" directive, followed by blank line

VandanaShah-Cornell:
My apologies for these errors, looks like I was testing the query without creating a table, and then sent the version I was mucking around with as a PR, instead of the cleaned version.

VandanaShah-Cornell:
Hi Irina,

I am so sorry about this, looks like I was testing the query without creating a table, and then sent version I was mucking around with as a PR, instead of the cleaned version.

-Vandana

VandanaShah-Cornell:
Re: Using both holdings_record__t and holdings_record tables : most of the fields are already extracted from the jsonb and are in the holdings_record__t table, except for three fields that needed to be extracted from the holdings_record table. It is easier to use the transformed table, is there any reason to not do this? Seems like extra work to extract fields that are already in transformed tables.

trapido:
Thank you very much, Vandana! Everything looks good. Just one minor thing: should the metadb dirctive read "--metadb:table holdings_ext" instead of "--Metadb: holdings_ext"? 
Regarding using holdings_record__t and holdings_record tables in the same query, there is nothing wrong with this, I just remember Nassib mentioning that this is not efficient because we are essentially joining two copies of the same data.

Issue #877, opened 2025-03-13
finance_budgets derived table for metadb missing in v1.7.12


sbeltaine:
The finance_budgets derived table is missing in v1.7.12 for metadb

PR #881, opened 2026-02-08
Add finance_budgets derived table for Metadb #877

sohamjana706-max:
Description: This Pull Request addresses issue #877 by adding the missing finance_budgets derived table specifically for the Metadb platform.

Changes Made:

New Derived Table: Created sql_metadb/derived_tables/finance_budgets.sql by porting the logic from the existing LDP version to Metadb-compliant syntax.

Metadb Directives: Added the required --metadb:table finance_budgets directive at the top of the SQL file.

Execution Order: Updated sql_metadb/derived_tables/runlist.txt to include the new table, ensuring it is built during the automated deployment process.

Syntax Standards: Followed repository guidelines regarding SQL keyword capitalization and column aliasing.

Verification:

Verified that the column names and calculation logic (COALESCE, CASE statements) match the original finance_budgets logic found in the LDP directory.

### DIFF
--- a/sql_metadb/derived_tables/finance_budgets.sql
+++ b/sql_metadb/derived_tables/finance_budgets.sql
@@ -3,85 +3,40 @@
 -- This derived table shows the budgets for each fund for all fiscal years
 -- including calculated values.
 
-DROP TABLE IF EXISTS finance_budgets;
-
-CREATE TABLE finance_budgets AS
 SELECT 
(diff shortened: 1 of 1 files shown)


hutchinsonca2:
The finance_budgets.sql table was included in v.1.8.0 & v.1.8.1.  I no longer think we need this PR.  Thanks.

